PR #4, opened 2016-05-18
Fix a race condition in setting the artwork

fastred:
#### What's this PR do?

It was possible to see a wrong artwork if events occurred in the following order:
1. Set now playing info to Track A
2. Download of artwork for Track A starts
3. Set now playing info to Track B
4. Download of artwork for Track A finishes and is set in nowPlayingInfo

#### How to test:

Please try to reproduce steps described above. You can add a breakpoint add this line:
```
if ([url isEqual:self.artworkUrl]) {
```

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -6,6 +6,12 @@
 
 @import MediaPlayer;
 
+@interface MusicControlManager ()
+
+@property (nonatomic, copy) NSString *artworkUrl;
+
+@end
+
 @implementation MusicControlManager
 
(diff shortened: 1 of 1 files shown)


tanguyantoine:
I will look at that quickly. :) so happy to have contribution

fastred:
> I will look at that quickly. :) so happy to have contribution

Happy to help! And thanks for creating this library in the first place 😄.

tanguyantoine:
DO you have some knowledge in Android Programming (it is obscure for me ^^)

PR #6, opened 2016-05-18
Add support for skip backward and forward

fastred:
#### What's this PR do?

This PR adds support for skipping backward and forward on iOS.

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -23,8 +23,8 @@ var MusicControl = {
   resetNowPlaying: function(){
     NativeMusicControl.resetNowPlaying()
   },
-  enableControl: function(controlName, bool, options){
-    NativeMusicControl.enableControl(controlName, bool, options)
+  enableControl: function(controlName, bool, options = {}){
+    NativeMusicControl.enableControl(controlName, bool, options || {})
   },
   handleCommand: function(commandName){
     if(handlers[commandName]){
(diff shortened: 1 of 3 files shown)


tanguyantoine:
Can you add example on the README for this feature?

index.ios.js
```diff
@@ -23,8 +23,8 @@ var MusicControl = {
   resetNowPlaying: function(){
     NativeMusicControl.resetNowPlaying()
   },
-  enableControl: function(controlName, bool){
-    NativeMusicControl.enableControl(controlName, bool)
+  enableControl: function(controlName, bool, options){
```
tanguyantoine: `enableControl: function(controlName, bool, options = {}){``

adding default will prevent null exception on Objective-C side no?

index.ios.js
```diff
@@ -23,8 +23,8 @@ var MusicControl = {
...
+    NativeMusicControl.enableControl(controlName, bool, options)
```
tanguyantoine: probably here too, what do you think of `options || {}` ?

tanguyantoine:
Except what i commented all is a good idea

fastred:
@tanguyantoine Good points, thanks! I added two new commits.

tanguyantoine:
You rocks

Issue #1, opened 2016-05-13
Android support


tanguyantoine:
# Description

**Android developper needed**

Add behavior to display and control on Android
# Tasks
- [X] Display infos
- [ ] Handle remote control events
- [ ] Display cover artwork from Url/local file

cinder92:
👍

Guichaguri:
@tanguyantoine Is this project still active? I might work on the Android support doing something similar from what I did in [RemoteControls](https://github.com/shi11/RemoteControls/pull/32)

tanguyantoine:
Feel free to do has many PR as you want :)

cinder92:
@Guichaguri  i just have to implement some function in this merge feel free to collaborate :D

dininski:
@cinder92 are you still on this?

cinder92:
@dininski i'm on, but sadly i'm not an android expert, i just edit some functionality, it needs to implement the controls on notification... @Guichaguri are you here to help us?

Guichaguri:
@cinder92 Yes. There are two things that need to be implemented: MediaSession (changes lockscreen background and makes it work with smartwatches and cars), Notification (just a simple notification with media buttons).

For the API, Android supports a lot of options.
Just take a look at the new API implementation I did in [RemoteControls](https://github.com/shi11/RemoteControls/pull/32), it supports most of the options that Android provides.

We should start discussing API changes, if we want to support most of them.

cinder92:
@Guichaguri  yes that true, i saw that this plugin is not working anymore with api 16... this is about media session that is available since API lvl 21, i'm not an android developer but if you need help i can do somenthing...

Guichaguri:
@cinder92 Version compatibility is not a problem when we can use the official support library. The notification works in almost every version and MediaSession can be easily replaced.

Above API 21: MediaSession and Notification
Between API 14-21: RemoteControlClient and Notification
Below API 14: Notification only

For the API, here's all the properties that I was thinking to support:
```js
MusicControl.enableControl('nextTrack', true); // Enabled by default
MusicControl.enableControl('previousTrack', true); // Enabled by default
MusicControl.enableControl('play', true); // Enabled by default
MusicControl.enableControl('pause', true); // Enabled by default
MusicControl.enableControl('stop', false); // Disabled by default
MusicControl.enableControl('seekTo', false); // Disabled by default
MusicControl.enableControl('rate', false); // Disabled by default
[...]

PR #21, opened 2016-12-04
[Android] MediaSession support

Guichaguri:
This PR adds support to MediaSession, and makes the Notification work on top of it
It also uses the support library for the MediaSession and Notification, so they work in any Android version.
I also made changes to the gradle config to make it work properly

Related to #1


Guichaguri:
Do not merge it yet. I still haven't tested it in a real player, and also, I'm trying to figure out a easy way to install it on Android (maybe supporting `react-native link`?)

tanguyantoine:
It does not support linking ?

Guichaguri:
Everything should work with linking, but I don't know if it registers the package automatically. I'll see about it when I have some free time

tanguyantoine:
Let me know when you are done 👍

Guichaguri:
I think I'm done here.
Changes that you should be aware of:
* Added RN link support (take a look at [react-native-device-info](https://github.com/rebeccahughes/react-native-device-info) with you want to know how they did it in iOS)
* Added a new function: `setPlayback`. It is needed to reduce metadata updates in `setNowPlaying`. (for instance, in iOS, it will download the artwork everytime you want to make an update to playback data)
* Added 5 constants (STATE_ERROR, STATE_STOPPED, STATE_PLAYING, STATE_PAUSED, STATE_BUFFERING) to use with the `state` parameter in `setPlayback`

tanguyantoine:
Awesome. 🎉

PR #29, opened 2017-01-15
Update iOS headers for RN 0.40

smkhalsa:


tanguyantoine:
What do you think of upgrading react-native version in `package.json` as well ?

smkhalsa:
@tanguyantoine I've updated the peer dependency and the npm version in package.json and removed the search paths.

Issue #25, opened 2016-12-21
Android: Volume buttons unresponsive


rikochet:
During playback volume buttons are unresponsive.

If I replace `session.setPlaybackToRemote()` with `session.setPlaybackToLocal()` and use Android's `AudioManager` I can circumvent this issue.

If I'm honest don't actually know the purpose of `setPlaybackToRemote()` .. Is it something to do with `MediaRouter`?

tanguyantoine:
Sorry I'm not an Android developer, @Guichaguri any thoughts ?

Guichaguri:
The reason why I use `setPlaybackToRemote` is to make it work with chromecast and other external devices.
It might be a good idea to also add an option to use `setPlaybackToLocal`...
I'll take a look when I have time :)

Guichaguri:
The remote volume has been fixed and there's a `remoteVolume` option which is disabled by default

Issue #23, opened 2016-12-20
Toggling "play" also starts the SoundCloud app


jasongrishkoff:
Real Android device (Nexus 5x w/ Android 7.1.1)

If I pause my song using:
`MusicControl.on('pause',` ()=> { this.playPause() })`

Firing the app up again with the following also causes my SoundCloud app to start playing again:
`MusicControl.on('play',` ()=> { this.playPause() })`

I'm wondering if they both use ExoPlayer and it's somehow linked to that?

rikochet:
I'm experiencing a similar issue..

The first time I play/pause VLC will start playing, and the second time Spotify will start.

Looking through the docs I thought [getSessionToken](https://github.com/tanguyantoine/react-native-music-control/blob/a8d22e324700686d84e78615b292ee136b45e817/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java#L83) was supposed to protect against this..

AshkanValadkhani:
same issue here ( same as @rikochet  ) and volume button not working

jasongrishkoff:
Yeah, volume button doesn't work for me, but I had assumed that was related to another issue and was still planning to look into it.  Looks like perhaps it's related to MusicControl.  I'll dig deeper, and if so include it in another open issue :)

jasongrishkoff:
Update: first resume after pause starts SoundCloud; if I pause and then play a second time, it launches Spotify.

rikochet:
@AshkanValadkhani 

I don't fully understand what `setPlaybackToRemote` does, so I just swapped it out for `setPlaybackToLocal` and used AudioManager, now the audio controls work fine.

Example;
```session.setPlaybackToLocal(AudioManager.STREAM_MUSIC);```

ahmedlhanafy:
I have this one as well pressing playing triggers all of music apps to play

jasongrishkoff:
A bit stumped on this one right now -- anyone else figure it out?

rikochet:
I wasted most of my christmas holiday trying to figure out what was causing this issue and eventually came to the conclusion that issue is with [session.getController().dispatchMediaButtonEvent(ke);](https://github.com/tanguyantoine/react-native-music-control/blob/master/android/src/main/java/com/tanguyantoine/react/MusicControlReceiver.java#L30)..

[...]

PR #33, opened 2017-01-19
Fixed volume and notification buttons

Guichaguri:
This PR fixes the volume controls (#25), fixes the notification buttons (#23) and adds two new options:

* `remoteVolume` in `enableControl` - Specifies whether the volume control is local or remote.
* `maxVolume` in `setPlayback` - Sets the maximum volume. Only works when `remoteVolume` is enabled.

Local volume control is automatic, which means you will not receive volume events or be able to disable the volume control. `volume` and `maxVolume` also doesn't do nothing

Remote volume is 100% up to you, you will receive volume events and be able to toggle the volume control but you have to specify the current volume and the maximum volume

The notification buttons now handle special actions that change depending on your app name, so no more conflict with other apps

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,3 +1,8 @@
-<manifest xml:android="http://schemas.android.com/apk/res/android"
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="com.tanguyantoine.MusicControl">
+
+    <application>
+        <service android:stopWithTask="false" android:name="com.tanguyantoine.react.MusicControlNotification$NotificationService" />
+    </application>
+
 </manifest>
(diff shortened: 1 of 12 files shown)


Guichaguri:
Note: `remoteVolume` will have the chromecast icon for the volume bar. This might not be what you want, but there's no way to change it unless you disable `remoteVolume`

android/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,7 @@
           package="com.tanguyantoine.react">
 
     <application>
-        <service android:stopWithTask="false" android:name="com.tanguyantoine.react.MusicControlNotification$NotificationService" />
```
tanguyantoine: :+1:

Issue #34, opened 2017-01-23
Android: Error using `setNowPlaying` without `rating`


shshaw:
# Description

1. Sample code (provide repo url or sample code)

```
MusicControl.setNowPlaying({ 
        artist: this.state.artist, 
        title: this.state.title,
        artwork: this.state.artwork || '',
        color: 0x000000
});
```
Produces this error: 
![screen shot 2017-01-23 at 11 56 02 am](https://cloud.githubusercontent.com/assets/777155/22216041/1929f6ee-e163-11e6-8ac2-1ca2c01f392b.png)

Adding `rating: 100` to the params makes it work without any trouble, though it seems like that shouldn't be required.

2. Platform ?
   - [ ] iOS
   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [ ] Real device

Guichaguri:
Okay, this is a little strange
The source code that comes with the gradle dependency can receive a null value, but the [real code](https://github.com/android/platform_frameworks_support/blob/master/media-compat/java/android/support/v4/media/MediaMetadataCompat.java#L803-L809) actually needs a non-null value for Android KitKat

PR #35, opened 2017-01-23
Fix not provided rating

tanguyantoine:
#### What's this PR do?

Fix error when rating is not provided

#34 

@Guichaguri do you think it can do the trick ?


Guichaguri:
This is actually the right fix for it, but would not work, the `newUnratedRating` needs a type as a parameter.
I fixed it in 72c521ce61ab15fe566cd312ecfc9b32a29b5954, it should tell the MediaSession that it accepts a rating if it's enabled, or tell it that it doesn't accept ratings otherwise

Issue #40, opened 2017-02-26
Pause does not strop duration


deshiknaves:
On RN 41. Everything works as expected. Except when I pause the audio, the duration doesn't pause - it keeps going one. Is there a way to manually pause the duration on iOS?

1. Sample code (provide repo url or sample code)

2. Platform ?
   - [X ] iOS
   - [ ] Android
   
   
3. Device
  - [X ] Simulator
  - [ ] Real device

deshiknaves:
Looked thought the source code and there is nothing currently in there to achieve this. I'll send through a pull request that will allow this.

tanguyantoine:
closed by #41

Issue #36, opened 2017-01-25
Add setPlayback to iOS


Guichaguri:
Currently, setPlayback is only available in Android. Would be nice to have support this function in iOS devices.
Options that can't be implemented in iOS should be ignored and a note should be included in the README clarifying what is Android-only and what is iOS-only

```js
MusicControl.setPlayback({
  state: MusicControl.STATE_PLAYING, // (STATE_ERROR, STATE_STOPPED, STATE_PLAYING, STATE_PAUSED, STATE_BUFFERING)
  volume: 10, // (Integer from 0 to maxVolume)
  maxVolume: 10, // (Integer)
  speed: 1, // Playback Rate (Double)
  elapsedTime: 103, // Elapsed Seconds (Double)
  bufferedTime: 200 // Buffered Seconds (Double)
})
```

tanguyantoine:
* speed: playbackRate
* elapsedTime: elapsedPlaybackTime
* state: what is that referring to ?
* volume/maxVolume : looks to not be customizable in iOs

Guichaguri:
Wouldn't it be better to move `playbackRate` and `elapsedPlaybackTime` to match the Android side? I've created the `setPlayback` function is for performance reasons, so you wouldn't have to update the metadata every time you just wanted to update the state.

`state` is whether the player is playing, paused, stopped, buffering, etc. Unfortunately, I don't think it's supported in iOS, I've only found an option for [macOS](https://developer.apple.com/reference/mediaplayer/mpnowplayingplaybackstate?language=objc)

Any option that is not customizable in iOS should be ignored

Guichaguri:
Added by #47

PR #41, opened 2017-02-27
Toggle pause method for iOS

deshiknaves:
#### What's this PR do?
Adds a new method to iOS so that the duration bar can be paused when pause is pressed. Currently when pause is hit, the duration bar keeps going forward. 

[#40](https://github.com/tanguyantoine/react-native-music-control/issues/40)

#### How to test:
Toggle pause after the playback has started. It will stop the duration bar and set the stopped time to the elapsed time. All other details of the info will remain what it was before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,8 @@ MusicControl.enableControl('enableLanguageOption', false); // iOS only
 MusicControl.enableControl('disableLanguageOption', false); // iOS only
 MusicControl.enableControl('skipForward', false); // iOS only
 MusicControl.enableControl('skipBackward', false); // iOS only
+
+MusicControl.togglePause(45); // Elapsed time in seconds, iOS only
 ```
 
 `skipBackward` and `skipForward` controls on iOS accept additional configuration options with `interval` key:
--- a/index.ios.js
+++ b/index.ios.js
@@ -14,6 +14,9 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
+  togglePause: function(elapsed) {
+    NativeMusicControl.togglePause(elapsed);
+  },
   enableBackgroundMode: function(enable){
     NativeMusicControl.enableBackgroundMode(enable)
   },
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -23,6 +23,27 @@ RCT_EXPORT_MODULE()
     return dispatch_get_main_queue();
 }
 
+RCT_EXPORT_METHOD(togglePause:(nonnull NSNumber *) elapsed)
+{
+    MPNowPlayingInfoCenter *center = [MPNowPlayingInfoCenter defaultCenter];
+    
+    if (center.nowPlayingInfo == nil) {
+        return;
+    }
+    
+    NSMutableDictionary *mediaDict = [[NSMutableDictionary alloc] initWithDictionary: center.nowPlayingInfo];
+
+    NSNumber *speed = [[mediaDict objectForKey:MPNowPlayingInfoPropertyPlaybackRate] isEqual:[NSNumber numberWithDouble:0]]
+        ? [NSNumber numberWithDouble:1]
+        : [NSNumber numberWithDouble:0];
+
+    
+    [mediaDict setValue:speed forKey:MPNowPlayingInfoPropertyPlaybackRate];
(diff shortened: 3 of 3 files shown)


Guichaguri:
Shouldn't this be done using the current player state?
Android already has the `setPlayback` which receives some useful information about the playback, including the current state (paused/playing/buffering/stopped/etc). ( #36 )
Android and iOS shouldn't have different API functions that produces the same result.

I don't know, but `togglePause` seems like a bad design choise for me...

deshiknaves:
Not everything updates this. React Native Audio does, but it doesn't support web audio. For example https://github.com/react-native-community/react-native-video uses AVAudioSession and can playback audo steams. This does not update the play/pause state of the global player.

Without this method whenever paused, the duration keeping going forward. I don't think this is an issue on Android.

deshiknaves:
The APIs should be aligned. However, needing to set all the other properties when you just want to toggle the paused/play state is overkill. For example play/pause has nothing to do with setting the volume.

I personally think they should both be included in both APIs.

deshiknaves:
Btw, if you guys like, I can implement the `setPlayback` method on iOS.

tanguyantoine:
@deshiknaves it would be awesome.

deshiknaves:
Looking at the source code for the Android version and iOS version of Media Center Manager, they both behave differently. The Android version `setNowPlaying` creates a new dictionary of metadata and sets it. The iOS version either creates a new one or it updates the existing one. So the `setNowPlaying` on iOS can be used to update one single key, which is not the case on Android.

The documentation doesn't really explain this difference or that you can pass in only a few properties to be set. I think this should probably be explained in more detail.

`setNowPlaying` sounds like it is going create a new Now Playing item, which would be consistent with the Android side of things.

So a lot of what what `setPlayback` does on the Android side is already covered in this method. If it's just for API consistency, then I add a method that achieves that. In that case, shouldn't the `setNowPlaying` be changed for consistency?

Let me know what you guys want to do here and I can implement it, or you guys can decided what you want to do with it later. 

Speaking to the `togglePause` method - this is a very common use case and having a method that does exactly that is more handy than having to set a whole bunch of other properties.

tanguyantoine:
I think having both will make everyone happy. 
This way we will be able to manage it differently if needed.
As a shortcut with `setNowPlaying` and as a specific case with what you already did. 
@Guichaguri what do you think ?

Guichaguri:
What if we change the `setPlayback` to just update the properties instead of recreating the playback metadata? This way, we can only specify the new state without updating the rest of the properties. No need for another platform dependent function.

deshiknaves:
Yep, that sounds like a good plan. Then should it then be able to update any property?

Also, should `setNowPlaying` on iOS reflect the behavior of Android?

Guichaguri:
Yep, `MusicControl.setPlayback({state: MusicControl.STATE_PAUSED});` would work without reseting volume, speed, etc

Currently, `setNowPlaying` recreates the metadata completely, but that can also be changed if needed. Any unknown property is also ignored.

deshiknaves:
The iOS version doesn't recreate it, it mutates what's already there. Which is completely valid. But whichever behavior we prefer both should probably do the same thing.

Yep, unknown properties are ignored in iOS as well.

Guichaguri:
I feel like `setNowPlaying` should reset, as you only will call this function when the track changes and you don't want to mix it with the info of the previous track.
`setPlayback` should not reset, as the player will be always the same even through different tracks.

This is what I think makes sense. What do you think?

deshiknaves:
Yep, I completely agree with you. I think that's what the method signature alludes to. I'll make those changes tomorrow and send through a PR.

It may have been nice if `setPlayback` was named `updatePlayback` but I think that's a moot point, the API is already out there :).

Guichaguri:
We can rename it, but we should keep the old one for compatibility

tanguyantoine:
@Guichaguri otherwise we can update the version and mark is as 'Breaking changes' :)

Guichaguri:
I think the Android part is done, I'll test it and I'll push it tomorrow.

deshiknaves:
I've just finished it too. I'll test it tonight and send through a PR.

tanguyantoine:
do you want me to publish the new version ?

Guichaguri:
@tanguyantoine let's finish first the `updatePlayback` and the `require` for local images to create only one "breaking changes" version

Issue #43, opened 2017-03-03
Android "require()" image support


jman48:
# Description
Android version does not support reactjs style require keyword to pass images as the artwork. If you use the require keyword to load an image it will throw IO error as music-control loads from file paths, not drawables. Will then just show the default play icon.

1. Sample code (provide repo url or sample code)
`MusicControl.setNowPlaying({ artwork: require('../images/album.png') });`

2. Platform ?
   - [ ] iOS
   - [X] Android
   
   
3. Device
  - [X] Simulator
  - [X] Real device

Guichaguri:
That would be definitely useful.
@tanguyantoine does the iOS version already supports it?

tanguyantoine:
kind of. In the iOS implementation it seems to rely on local path and you do not need to use the require function : https://github.com/tanguyantoine/react-native-music-control/blob/master/ios/MusicControlManager.m#L251

Could be great to abstract that a lil bit more.

How do you want to call this api

```js
MusicControl.setNowPlaying({ artwork: require('../images/album.png') });
```
or
```js
MusicControl.setNowPlaying({ artwork: 'images/album.png' });
```

The first one is probably clearer in a sense of if it is a string then it is an url otherwise it is a local file.

What do you think ?

Guichaguri:
@tanguyantoine I agree, the `require` one fits better with React Native, as other local image resources are also specified with `require`.

tanguyantoine:
created #46

PR #44, opened 2017-03-03
Support for "require('./image.png)" keyword for artwork

jman48:
#### What's this PR do?
It will allow users to use the require keyword to load images for album artwork. e.g 
`MusicControl.setNowPlaying({ artwork: require('../img/album.png') });`
Android only.

https://github.com/tanguyantoine/react-native-music-control/issues/43

#### How to test:
Use the require keyword to load an image and check if it shows as the album artwork

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -18,6 +18,7 @@ import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
+import com.facebook.react.bridge.ReadableType;
 import com.facebook.react.views.imagehelper.ResourceDrawableIdHelper;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -43,6 +46,8 @@
 
     private boolean isPlaying = false;
     private long controls = 0;
+    //Flag to check if artwork is passed using react style "require('./image.png)"
+    private boolean fromRequire = false;
```
Guichaguri: Why create another field when you can just pass as a method parameter?

index.android.js
```diff
@@ -22,6 +23,12 @@ var MusicControl = {
     NativeMusicControl.enableBackgroundMode(enable)
   },
   setNowPlaying: function(info){
+    //Check if we have an android asset id from react style image "require('./image.png')"
+    if (info.artwork && !isNaN(info.artwork)) {
+      info.artwork = resolveAssetSource(info.artwork).uri;
```
Guichaguri: Wouldn't be simpler if this was handled in the Android side? You just need to check if it's a ReadableMap and get the `uri` from it.

jman48: We have to resolve the asset on the JS side as the resolveAssetSource function does not have a java equivalent. If I understand you we should remove the ".uri" and then check on java side if it is a map and extract uri that way?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
jman48: Yeah good call. Will get it changed

index.android.js
Guichaguri: Yep, that seems like a better solution over `fromRequire`

jman48: Latest changes will now detect and extract URI on android side

Guichaguri:
Thank you :)

tanguyantoine:
I would have prefer to wait for iOS implementation but too late :)

PR #47, opened 2017-03-04
Added updatePlayback and setPlayback. Some refactoring.

deshiknaves:
#### What's this PR do?
* Changed the `setNowPlaying` to always use a new dictionary.
* Refactored updating of the dictionary using a loop and a keys dictionary. Should do this for the actions too.
* Added `updatePlayback` which updates the current media data with any keys passed in the info object.
* Added `setPlayback` which using `updatePlayback` for backwards compatibility. We can choose to remove this.
* Removed `togglePause` as this is not needed if there is an `updatePlayback` method.

[#36](https://github.com/tanguyantoine/react-native-music-control/issues/36)
[#40](https://github.com/tanguyantoine/react-native-music-control/issues/40)

#### How to test:
* Every time `setNowPlaying` is called it should create a new playback.
* `updatePlayback` should be able to update all known keys

Let me know if I should change anything in here. Also if anything needs to change for the Android and iOS APIs to align.

### DIFF
--- a/index.ios.js
+++ b/index.ios.js
@@ -14,6 +14,9 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
+  togglePause: function(elapsed) {
+    NativeMusicControl.togglePause(elapsed);
+  },
   enableBackgroundMode: function(enable){
     NativeMusicControl.enableBackgroundMode(enable)
   },
--- a/index.ios.js
+++ b/index.ios.js
@@ -14,6 +14,8 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
+  STATE_PLAYING: NativeMusicControl.STATE_PLAYING,
+  STATE_PAUSED: NativeMusicControl.STATE_PAUSED,
   setPlayback: function (info) {
     NativeMusicControl.updatePlayback(info);
   },
--- a/index.ios.js
+++ b/index.ios.js
@@ -14,8 +14,8 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
-  togglePause: function(elapsed) {
-    NativeMusicControl.togglePause(elapsed);
+  updatePlaying: function(info) {
+    NativeMusicControl.updatePlaying(info);
   },
   enableBackgroundMode: function(enable){
     NativeMusicControl.enableBackgroundMode(enable)
--- a/index.ios.js
+++ b/index.ios.js
@@ -16,6 +16,9 @@ var subscription = null;
 var MusicControl = {
   STATE_PLAYING: NativeMusicControl.STATE_PLAYING,
   STATE_PAUSED: NativeMusicControl.STATE_PAUSED,
+  STATE_ERROR: NativeMusicControl.STATE_PAUSED,
+  STATE_STOPPED: NativeMusicControl.STATE_PAUSED,
+  STATE_BUFFERING: NativeMusicControl.STATE_PAUSED
   setPlayback: function (info) {
     NativeMusicControl.updatePlayback(info);
   },
--- a/index.ios.js
+++ b/index.ios.js
@@ -14,8 +14,11 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
-  updatePlaying: function(info) {
-    NativeMusicControl.updatePlaying(info);
+  setPlayback: function (info) {
+    NativeMusicControl.updatePlayback(info);
+  },
+  updatePlayback: function(info) {
+    NativeMusicControl.updatePlayback(info);
   },
   enableBackgroundMode: function(enable){
     NativeMusicControl.enableBackgroundMode(enable)
--- a/index.ios.js
+++ b/index.ios.js
@@ -18,7 +18,7 @@ var MusicControl = {
   STATE_PAUSED: NativeMusicControl.STATE_PAUSED,
   STATE_ERROR: NativeMusicControl.STATE_PAUSED,
   STATE_STOPPED: NativeMusicControl.STATE_PAUSED,
-  STATE_BUFFERING: NativeMusicControl.STATE_PAUSED
+  STATE_BUFFERING: NativeMusicControl.STATE_PAUSED,
   setPlayback: function (info) {
     NativeMusicControl.updatePlayback(info);
   },
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -23,6 +23,27 @@ RCT_EXPORT_MODULE()
     return dispatch_get_main_queue();
 }
 
+RCT_EXPORT_METHOD(togglePause:(nonnull NSNumber *) elapsed)
+{
+    MPNowPlayingInfoCenter *center = [MPNowPlayingInfoCenter defaultCenter];
+    
+    if (center.nowPlayingInfo == nil) {
+        return;
+    }
+    
+    NSMutableDictionary *mediaDict = [[NSMutableDictionary alloc] initWithDictionary: center.nowPlayingInfo];
+
+    NSNumber *speed = [[mediaDict objectForKey:MPNowPlayingInfoPropertyPlaybackRate] isEqual:[NSNumber numberWithDouble:0]]
+        ? [NSNumber numberWithDouble:1]
+        : [NSNumber numberWithDouble:0];
+
+    
+    [mediaDict setValue:speed forKey:MPNowPlayingInfoPropertyPlaybackRate];
+    [mediaDict setValue:elapsed forKey:MPNowPlayingInfoPropertyElapsedPlaybackTime];
+    
+    center.nowPlayingInfo = mediaDict;
+}
+
 RCT_EXPORT_METHOD(setNowPlaying:(NSDictionary *) details)
 {
 
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -9,8 +9,11 @@
 @interface MusicControlManager ()
 
 @property (nonatomic, copy) NSString *artworkUrl;
+@property (nonatomic, strong) NSDictionary *mediaKeys;
 
 @end
+
+NSString * const MEDIA_PLAYBACK_RATE = @"playbackRate";
 
 @implementation MusicControlManager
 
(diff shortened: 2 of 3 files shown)


index.ios.js
```diff
@@ -14,8 +14,8 @@ var handlers = { };
 var subscription = null;
 
 var MusicControl = {
-  togglePause: function(elapsed) {
-    NativeMusicControl.togglePause(elapsed);
+  updatePlaying: function(info) {
```
tanguyantoine: `updatePlayback`

ios/MusicControlManager.m
```diff
@@ -9,21 +9,56 @@
...
 
+NSString * const MEDIA_PLAYBACK_RATE = @"playbackRate";
+
 @implementation MusicControlManager
 
 @synthesize bridge = _bridge;
 
 RCT_EXPORT_MODULE()
 
+- (NSDictionary *) mediaKeys {
```
tanguyantoine: What do you think of defining that as an static const ?

ios/MusicControlManager.m
```diff
@@ -193,10 +156,27 @@ - (dispatch_queue_t)methodQueue
...
 
+- (NSDictionary *) update:(NSMutableDictionary *) mediaDict with:(NSDictionary *) details andSetDefaults:(BOOL) setDefault {
+    
+    for (NSString *key in [self mediaKeys]) {
+        if ([details objectForKey:key] != nil) {
+            [mediaDict setValue:[details objectForKey:key] forKey:[[self mediaKeys] objectForKey:key]];
+        }
+        
+        // In iOS Simulator, always include the MPNowPlayingInfoPropertyPlaybackRate key in your nowPlayingInfo dictionary
+        // only if we are creating a new dictionalty
```
tanguyantoine: do you mean `Dictionary` ?

ios/MusicControlManager.m
```diff
@@ -9,21 +9,56 @@
 @interface MusicControlManager ()
 
 @property (nonatomic, copy) NSString *artworkUrl;
+@property (nonatomic, strong) NSDictionary *mediaKeys;
 
 @end
 
+NSString * const MEDIA_PLAYBACK_RATE = @"playbackRate";
+
```
tanguyantoine: Would you mind trying with this piece of code ? Do you think it could work ?

```objective-c
#define mediaKeys  @{  @"album": MPMediaItemPropertyAlbumTitle, \
    @"trackCount": MPMediaItemPropertyAlbumTrackCount, \
    @"trackNumber": MPMediaItemPropertyAlbumTrackNumber, \
    @"artist": MPMediaItemPropertyArtist, \
    @"composer": MPMediaItemPropertyComposer, \
    @"discCount": MPMediaItemPropertyDiscCount, \
    @"discNumber": MPMediaItemPropertyDiscNumber, \
    @"genre": MPMediaItemPropertyGenre, \
    @"persistentID": MPMediaItemPropertyPersistentID, \
    @"duration": MPMediaItemPropertyPlaybackDuration, \
    @"title": MPMediaItemPropertyTitle, \
    @"elapsedPlaybackTime": MPNowPlayingInfoPropertyElapsedPlaybackTime, \
    @"playbackRate": MPNowPlayingInfoPropertyPlaybackRate, \
    @"playbackQueueIndex": MPNowPlayingInfoPropertyPlaybackQueueIndex, \
    @"playbackQueueCount": MPNowPlayingInfoPropertyPlaybackQueueCount, \
    @"chapterNumber": MPNowPlayingInfoPropertyChapterNumber, \
    @"chapterCount": MPNowPlayingInfoPropertyChapterCount \
}
```

deshiknaves: You can't define a const that is a dictionary. I can make it a static dictionary though. I'll make that change in the morning and create a new PR.

ios/MusicControlManager.m
deshiknaves: Yep, I'll update that.

index.ios.js
deshiknaves: Whoops. Should be. I'll update it.

ios/MusicControlManager.m
deshiknaves: It can't be a const, but it can be static. I'll change it.

ios/MusicControlManager.m
```diff
@@ -9,21 +9,56 @@
...
+                @"artist": MPMediaItemPropertyArtist,
+                @"composer": MPMediaItemPropertyComposer,
+                @"discCount": MPMediaItemPropertyDiscCount,
+                @"discNumber": MPMediaItemPropertyDiscNumber,
+                @"genre": MPMediaItemPropertyGenre,
+                @"persistentID": MPMediaItemPropertyPersistentID,
+                @"duration": MPMediaItemPropertyPlaybackDuration,
+                @"title": MPMediaItemPropertyTitle,
+                @"elapsedPlaybackTime": MPNowPlayingInfoPropertyElapsedPlaybackTime,
+                @"playbackRate": MPNowPlayingInfoPropertyPlaybackRate,
```
Guichaguri: I don't know if this is supposed to work like this, but in Android these properties are called `speed` and `elapsedTime` instead of `playbackRate` and `elapsedPlaybackTime`.
Also, where is the `state`? Isn't it supposed to replace `togglePause`?

deshiknaves: There is no state on iOS. These are they keys that have always been here. If we are changing them, then it's a massive breaking change.

deshiknaves:
I've pushed some new changes.

ios/MusicControlManager.m
tanguyantoine: Why can't it be a const ?

tanguyantoine: if it always result to the same result it must be a constant IMHO

ios/MusicControlManager.m
deshiknaves: @tanguyantoine happy for `elapsedPlaybackTime` to change to `elapsedTime` and `playbackRate` to `speed`?

ios/MusicControlManager.m
tanguyantoine: I tried in XCode and no error was raised. That's why I'm asking that. At least can you make the method at the classe level instead of the instance level.

ios/MusicControlManager.m
tanguyantoine: Not objection agains that

ios/MusicControlManager.m
deshiknaves: Actually you're right, if you have the backwards slashes it will allow it.

deshiknaves: Btw static is class level not instance level.

deshiknaves:
Those changes have been made

ios/MusicControlManager.m
deshiknaves: Change has been pushed.

ios/MusicControlManager.m
deshiknaves: Took out the backslashes the first time

tanguyantoine:
Looks good. Did not had time to test locally. You can merge :)

ios/MusicControlManager.m
Guichaguri: Wasn't the state going to be used to toggle the pause button?
`STATE_PLAYING`: Shows the pause button
`STATE_PAUSED`: Shows the play button

deshiknaves: That doesn't exist on iOS. PlaybackRate 0 is pause 1 is play. You can see your comments here [#36](https://github.com/tanguyantoine/react-native-music-control/issues/36).

Guichaguri: Can you change that so it matches the Android side? It's a little confusing having to not only change the state, but also change the speed
You should be able to store the state value, overwriting the speed when it's paused...

deshiknaves: Can do. I'll do it in the morning. I'm out for the day.

deshiknaves: On iOS you have to specify the elapsed time otherwise it will reset to 0. You also can't retrieve the elapsed time. It just doesn't work the same way as Android. I can change it so it takes a symbol and the set the playback rate accordingly, but the user will have to provide the elapsed time. Also if a user uses both speed and state - which one takes precedence?

Guichaguri: You can store the stuff you need outside the dictionary, right?
When the state is set to paused, the speed has to be set to 0.
When the state is set back to playing, the speed can return back to what it used to be.

deshiknaves: There is no elapsed time to store. It starts counting up from the time you give it. When you pause you have to tell it at what time to pause. That time comes from else where - the state of your playback outside the displayed info.

It's not about storing the information. I'm asking about the precedence. If I supply { state: STATE_PLAYING, speed: 0 } which one wins?

deshiknaves: I also understand that on Android speed can be other than 0 and 1.

Guichaguri: In this case, speed. Only when the state is paused the state takes precedence

deshiknaves: You can't get `state.getPosition()` on iOS. MP does not keep track of position, that is to be sent in from whatever is playing the audio on iOS. It makes a lot more sense to not calculate the time in 2 different places and instead let people know in the docs that they need to pass in the elapsed time. That way the elapsed time is always the time that the audio playback has determined.

In your above example - if `STATE_PLAYING` and the user sends `0` as `speed`. It will be the same as `STATE_PAUSED`.

deshiknaves: Just to be clear. I can easily calculate the elapsed time by keeping track of the system time and updated time. However, it isn't the responsibility of `nowPlayingInfo` to keep track of time, it's your audio player. For an audio app, it makes sense for the audio track to determine where it is up to in it's playback. This is the reason why on iOS it doesn't return back where the UI is showing where it is at. If you're changing it, you need to set where it is up to.

Guichaguri: Android doesn't keep track of the position either. `getPosition` returns the position from when it was set, just like I stored the position somewhere and I were retrieving. If it returned the actual position, I didn't had to calculate the new position by myself.

Setting the speed to 0 will indeed be the same as `STATE_PAUSED`, as well as it would in Android. This is a programming problem with the app using this package, not ours.

It's fine if you want to require the app dev to set the elapsed time, feel free to do so. I might also change in my code, so I don't need to calculate it anymore. In Android, the elapsed time is only used and calculated by third party apps/devices such as smartwatches, that's why we can't retrieve in Android either.

deshiknaves: Yep, looking at the docs I got that it was getting the position from when it was last set. iOS doesn't have that property, which we can achieve by setting it in out own data structure. Which would enable us to calculate it the way you are on Android when it is not provided.

I do think it will be closer to the real thing and more consistent if the current time from the player is sent though in all situations - making it the same as the real class we making a facade for though this code.

I'll implement the states in the next hour and let you know when it's done.

deshiknaves:
@Guichaguri @tanguyantoine I've added state to the `updatePlayback` method.

ios/MusicControlManager.m
Guichaguri: I agree, but I'll not remove the code from Android, as all calls to `updatePlayback` would require the `elapsedTime`. I'll probably leave a note that providing the elapsed time is not only better for the performance but also for iOS compatibility.

Ah, before I forget, for the other Android states, can you set them as aliases to `STATE_PAUSED` just to not break compatibility? Something like this:
```js
STATE_PLAYING: NativeMusicControl.STATE_PLAYING,
STATE_PAUSED: NativeMusicControl.STATE_PAUSED,
STATE_ERROR: NativeMusicControl.STATE_PAUSED,
STATE_STOPPED: NativeMusicControl.STATE_PAUSED,
STATE_BUFFERING: NativeMusicControl.STATE_PAUSED
```

Thanks :)

deshiknaves: Too easy. I'll update that now.

Guichaguri:
Looks good to me.
It's up to @tanguyantoine to check it and merge it :)

PR #56, opened 2017-04-03
Remove flow annotation

cmcewen:
This removes the `@flow` annotation from the JS files. If the annotation is used without actually adding types to the files, it will throw errors on other files that import this and use flow

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-music-control",
-  "version": "0.3.4",
+  "version": "0.4.1",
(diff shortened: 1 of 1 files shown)


tanguyantoine:
Would it be better to add flow typing instead ?

cmcewen:
yeah definitely IMO - this will just stop errors since you are telling the flow checker that this file is annotated when it isn't, but adding types would be the best move

tanguyantoine:
so let's do that  ;-)

PR #57, opened 2017-04-03
adds support for iOS audio routing notifications

mzmiric5:
#### What's this PR do?

This PR adds support for iOS Audio Routing notifications. On iOS when a user plugs out the headphones or a bluetooth headset disconnects, audio playback stops. The remote control UI indicates this, but the pause event is not emitted down to the react-native app. This PR adds a notification center listener, and just emits an onPause event when it detects that the headphones have been unplugged.

Not related to an issue, was just a specific use case I had, and I thought it would be a useful feature to contribute back into the project.

N/A

#### How to test:

To test, when setting up events in react-native code, call MusicControl.listenToAudioRouteChange(true); in componentDidMount of your player app and also listen to the MusicControl pause event. Run the app on a physical iOS device and plug in headphones. Play audio and then plug out the headphones. The pause event will fire.

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -113,7 +113,7 @@ RCT_EXPORT_METHOD(enableControl:(NSString *) controlName enabled:(BOOL) enabled 
     
     if ([controlName isEqual: @"pause"]) {
         [self toggleHandler:remoteCenter.pauseCommand withSelector:@selector(onPause:) enabled:enabled];
-        [self listenToAudioRouteChange:true];
+        [self listenToAudioRouteChange:enabled];
     } else if ([controlName isEqual: @"play"]) {
         [self toggleHandler:remoteCenter.playCommand withSelector:@selector(onPlay:) enabled:enabled];
         
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -57,32 +57,32 @@ RCT_EXPORT_MODULE()
 RCT_EXPORT_METHOD(updatePlayback:(NSDictionary *) originalDetails)
 {
     MPNowPlayingInfoCenter *center = [MPNowPlayingInfoCenter defaultCenter];
-    
+
     if (center.nowPlayingInfo == nil) {
         return;
     }
-    
+
     NSMutableDictionary *details = [originalDetails mutableCopy];
-    
+
     // Set the playback rate from the state if no speed has been defined
     // If they provide the speed, then use it
     if ([details objectForKey:MEDIA_STATE] != nil && [details objectForKey:MEDIA_SPEED] == nil) {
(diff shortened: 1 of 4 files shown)


Guichaguri:
Instead of adding a new function, why don't you just trigger the pause event?

mzmiric5:
I mean that is what it does. the event handler in the iOS code triggers the pause event. I've only added a new function to it (which can be used to toggle the functionality on and off) since some users might not want this feature. The OS will cause the playback to pause either way, but my train of thought was that it was up to the developer if they want their app to listen to this event or not. I'd be happy to refactor it to listen to this event by default without a function to toggle it on and off, if that is the preferred way you would like it to be integrated

Guichaguri:
If the app is forced to pause playback, I don't see why a dev would want to desync their app, but if you're going to implement it as optional, please, use `enableControl` instead of creating a new function

mzmiric5:
ok, I guess I'll just refactor it as non optional. I was following the function approach that was used for toggling background audio modes as that made more sense vs enable control, since it's not an actual control in the remote UI. anywho, let me just make the change and push that

mzmiric5:
@Guichaguri ok, actually I see your point about it being enabled through the control. I've changed it so that it's actually gets toggled with the pause control, since it ends up triggering that same event

ios/MusicControlManager.m
```diff
@@ -188,6 +188,14 @@ - (void) toggleHandler:(MPRemoteCommand *) command withSelector:(SEL) selector e
     command.enabled = enabled;
 }
 
+- (void) listenToAudioRouteChange:(BOOL) enabled {
```
tanguyantoine: what do you think of 
```objective-c
- (void) listenToAudioRouteChange:(BOOL) enabled {
  if (enabled) {
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(audioHardwareRouteChanged:) name:AVAudioSessionRouteChangeNotification object:nil];
  } else { 
   [[NSNotificationCenter defaultCenter] removeObserver:self name:AVAudioSessionRouteChangeNotification object:nil];
  }
}
```

mzmiric5: is this safe to do? I was worrying about the enable triggering multiple times for whatever reason (user error, or react lifecycle being triggered again) so I just followed the same approach you used for control toggles. If you think the else approach is still a good idea, and that this wouldn't keep adding listeners, I'm happy to make the change

tanguyantoine: I think it is safe IMO. If we run into multiple subscription it means that there is another problem somewhere else. 

Otherwise we can addObserver on init and removeObserver on dealloc. This way no need to worry about multiple trigger. what do you think ?

mzmiric5: oh yeah the init and dealloc route makes way more sense to me. Made that change and updated the README, let me know if this is all good

PR #61, opened 2017-04-19
Add tvOS target

cmcewen:
This adds a tvOS target so the project will compile on tvOS react native apps


tanguyantoine:
I'm looking forward to see a screenshot on tvOS 😁

PR #73, opened 2017-07-17
Android synchronize init and destroy

noymer:
#### What's this PR do?

* When `MusicControlModule.onTrimMemory()` is called and almost at the same time 
 `setNowPlaying` or `updatePlayback()` or `enableControl()` is called, 
in `destroy()` many properties become null, but `init()` will not be called because they are called from different threads.
 It will cause NullPointerException.
Example:
``` 
// line 268
state = pb.build();
```

* Fix it by making  `destroy()` and other methods `synchronized`.


Guichaguri:
Thanks

tanguyantoine:
published the new version 0.4.6

Issue #46, opened 2017-03-04
Support local image for artwork


tanguyantoine:
# Description

Expected api call : `MusicControl.setNowPlaying({ artwork: require('../images/album.png') }); `

same as #44 and #43  but for iOs 

1. Platform ?
   - [x] iOS

Guichaguri:
I'll leave here some information which might help you for iOS support for this.

First of all, you need to resolve the require, RN has a function for this which is located at `react-native/Libraries/Image/resolveAssetSource`. [Here](https://github.com/tanguyantoine/react-native-music-control/blob/master/index.android.js#L34-L36) is the JS code used in the Android side, you might want to copy it to the iOS index.

After you passed the result to the native function, you need to check the type of the "artwork" property:
If it's a map, you can assume it's local and the URI is located in the `uri` property (string) of the map.
If it's a string, you can assume it's remote and the URL is the string.

At least for Android, RN has a helper method for loading local image URIs. If you don't find anything, you can load them the way you used to load local images before.

SteffeyDev:
It appears that the android fix would work on iOS as well, because using 

`MusicControl.setNowPlaying({artwork: resolveAssetSource(require('path/to/image.png')).uri});`

fixed it for me on iOS.

Guichaguri:
@SteffeyDev It works but it's not the proper solution to this problem. Currently, the iOS side detects when the artwork starts with "http", if it does, it loads a remove artwork, if not, it loads a local one.
In Android, it always loads strings as remote artworks and `require` objects as local artworks. This is better because we can have remote URLs with other protocols, such as ftp.

SteffeyDev:
I am only guessing here, but you could just pass the result of resolveAssetSource || the string to the native code like in index.android.js, then in objective c test if you get a dictionary object.  If yes, use the dictionary["uri"], otherwise load the string as remote.

tanguyantoine:
> MusicControl.setNowPlaying({artwork: resolveAssetSource(require('path/to/image.png')).uri});

Does that work to display local image ? would be pretty straight forward to make it work on iOs this way :)

SteffeyDev:
That would certainly work, but for consistancy between iOS and Android you should just do in index.ios.js what you did in index.android.js, and then modify the iOS native module to do the detection and switching. I'll be working on the android native code to figure out why it's not working, so I can test this out as well and throw in a pull request if it works.

SteffeyDev:
[...]

PR #76, opened 2017-08-05
Artwork Fixes, Code Cleanup, Skip on Android, etc (Issue #46)

SteffeyDev:
#### What's this PR do?
This PR is all of the changes I needed to make in order to use this library in my app.  This includes adding support for skip on android (including new images) and making the album artwork updates more consistent across platforms (supporting local and remote on simulator and real device).  In addition, I merged index.android.js and index.ios.js into musicControl.js and then specified this as the start file in package.json, which makes it a bit easier to act cross platform as the two versions were not really needed, and should make future changes involving that file easier.

Mainly #46, but may fix other ones with the various artwork fixes

#### How to test:
Test all usage (as if everything changed), because there are so many changes.  I will say that it is working OK in my app but I have unfortunately not been able to test it comprehensively.  At least test loading artwork on Android and iOS and using the skip option on both platforms


Guichaguri:
There are a lot of things to review here. First of all, @tanguyantoine what is `skipForward` and `skipBackward` supposed to do on iOS?

tanguyantoine:
@Guichaguri from what is implemented in the codebase it let you jump 10/30 (whatever interval) second forward or backward.

Merging `index.ios` and `index.android` could make sense if we ensure we can map functionnalities on both patform. Will review this more carefully

SteffeyDev:
@tanguyantoine The only difference between the two files previously where that enableControl accepted 3 arguments on iOS and 2 on Android.  The third options argument may be useful on android moving forwards, so while it is not needed now, the java was updated to allow options to be passed in.  This allowed the merge of the two nearly identical files.

SteffeyDev:
In case it's not clear from the code, all I did to enable skip on android was say that if one of the skip controls is enabled, use the skip images instead of the next/previous images and send the events to skipForward/skipBackward instead of nextTrack/previousTrack, that way skip can be used cross platform. I did update the readme to warn users not to mix track and skip controls, so that this is hopefully clear to them.

Guichaguri:
Your commit replaces the next/previous button to control it, which can lead to various problems.
You've only changed the notification, but the notification is not the only way to control the playback using this module, as you can use cars, smartwatches, headphone buttons, etc. All of those will still use the next/previous buttons which can lead to a bad UX.

Also, Android has a proper way to implement forward/rewind buttons, you should use it instead of replacing an existing button, as the dev may want both buttons to work.

For the other fixes and changes, you should split those into different commits, making it easier to review and merge

SteffeyDev:
I have split the changes into three commits based on features.

 I hadn't thought about the UX of the skip switch, you are completely right.  I will investigate to see if I can show the skip buttons only in the notification, and not broadcast the button availability to other user interfaces.  Let me know if you have a good idea of how to do this, or if this is even the correct approach.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -292,11 +301,20 @@ synchronized public void resetNowPlaying() {
     }
 
     @ReactMethod
-    synchronized public void enableControl(String control, boolean enable) {
-        init();
+    synchronized public void enableControl(String control, boolean enable, ReadableMap options) {
+        if(!init) init();
```
Guichaguri: `init` checks if it's not initialized, no need to check before invoking it.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -292,11 +301,20 @@ synchronized public void resetNowPlaying() {
...
 
         long controlValue;
+        boolean showSkip = false;
         switch(control) {
+            case "skipForward":
+                showSkip = true;
+                controlValue = PlaybackStateCompat.ACTION_SKIP_TO_NEXT;
```
Guichaguri: Don't use skip controls, use rewind/forward controls.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -270,6 +275,10 @@ synchronized public void updatePlayback(ReadableMap info) {
         state = pb.build();
         session.setPlaybackState(state);
 
+        if (info.hasKey("artwork")) {
```
Guichaguri: Keep the same code style

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -212,33 +203,47 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
-                }
-            });
-            artworkThread.start();
+        if (metadata.hasKey("artwork")) {
+          setArtwork(metadata);
         } else {
-            md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, null);
-            nb.setLargeIcon(null);
+          md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, null);
+          nb.setLargeIcon(null);
```
Guichaguri: Keep the same code style

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -212,33 +203,47 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
+              md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, bitmap);
+              nb.setLargeIcon(bitmap);
+
+              session.setMetadata(md.build());
+              notification.show(nb, isPlaying);
+              artworkThread = null;
+          }
+      });
+      artworkThread.start();
+    }
```
Guichaguri: Keep the same code style here too

android/src/main/java/com/tanguyantoine/react/MusicControlListener.java
```diff
@@ -51,11 +51,13 @@ public void onStop() {
     @Override
     public void onSkipToNext() {
         sendEvent(context, "nextTrack", null);
+        sendEvent(context, "skipForward", null);
     }
 
     @Override
     public void onSkipToPrevious() {
         sendEvent(context, "previousTrack", null);
+        sendEvent(context, "skipBackward", null);
```
Guichaguri: Again, use rewind/forward controls instead of next/previous controls

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -20,7 +20,7 @@
     private final ReactApplicationContext context;
 
     private int smallIcon;
-    private NotificationCompat.Action play, pause, stop, next, previous;
+    private NotificationCompat.Action play, pause, stop, next, previous, skipForward, skipBackward;
```
Guichaguri: I don't understand why you add new actions if you're replacing the next/previous actions

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -33,12 +33,12 @@ public MusicControlNotification(ReactApplicationContext context) {
...
 
-    public void updateActions(long mask) {
+    public void updateActions(long mask, boolean useSkip) {
         play = createAction("play", "Play", mask, PlaybackStateCompat.ACTION_PLAY, play);
         pause = createAction("pause", "Pause", mask, PlaybackStateCompat.ACTION_PAUSE, pause);
         stop = createAction("stop", "Stop", mask, PlaybackStateCompat.ACTION_STOP, stop);
-        next = createAction("next", "Next", mask, PlaybackStateCompat.ACTION_SKIP_TO_NEXT, next);
-        previous = createAction("previous", "Previous", mask, PlaybackStateCompat.ACTION_SKIP_TO_PREVIOUS, previous);
+        next = createAction(!useSkip ? "next" : "skip_forward", "Next", mask, PlaybackStateCompat.ACTION_SKIP_TO_NEXT, next);
+        previous = createAction(!useSkip ? "previous" : "skip_backward", "Previous", mask, PlaybackStateCompat.ACTION_SKIP_TO_PREVIOUS, previous);
```
Guichaguri: This would not always work, as the action would be already created previously.
Is there really a reason to add rewind/forward to the notification? Can't it just be available to external devices?

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -49,6 +49,8 @@ public void show(NotificationCompat.Builder builder, boolean isPlaying) {
         if(pause != null && isPlaying) builder.addAction(pause);
         if(stop != null) builder.addAction(stop);
         if(next != null) builder.addAction(next);
+        if(skipForward != null) builder.addAction(skipForward);
+        if(skipBackward != null) builder.addAction(skipBackward);
```
Guichaguri: If you're never initializing skipForward/skipBackward, why add them?

index.ios.js
```diff
@@ -1,77 +0,0 @@
...
-var handlers = { };
-var subscription = null;
-
-var MusicControl = {
-
-  STATE_PLAYING: NativeMusicControl.STATE_PLAYING,
-  STATE_PAUSED: NativeMusicControl.STATE_PAUSED,
-  STATE_ERROR: NativeMusicControl.STATE_PAUSED,
-  STATE_STOPPED: NativeMusicControl.STATE_PAUSED,
-  STATE_BUFFERING: NativeMusicControl.STATE_PAUSED,
```
Guichaguri: I don't think merging index.ios.js and index.android.js is a good idea. Have you tested everything to make sure it works?
For instance, the constants in iOS and Android are defined in a different way, as you can see here.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
Guichaguri: Also, is there a reason to move this to another method? It's only used here. I understand that is makes the code cleaner, but it counts towards the dex method limit.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -361,7 +379,8 @@ private Bitmap loadArtwork(String url, boolean local) {
         Bitmap bitmap = null;
 
         try {
-            if(local) {
+            // If we are running the app in debug mode, the "local" image will be served from htt://localhost:8080, so we need to check for this case and load those images from URL
+            if(local && !url.startsWith("http")) {
```
Guichaguri: This works, but it wouldn't work for other schemes. Take a look what I've done in my own module: https://github.com/Guichaguri/react-native-track-player/blob/dev/android/src/main/java/guichaguri/trackplayer/logic/track/Track.java#L171-L181

SteffeyDev:
Ok so I updated the skip controls so now it works as a stand in for seekForward and seekBackward, so app developers can choose to enable seekForward or skipForward and seekBackward or skipBackward, the only difference being the icon that shows up on the notification.  I believe this is a reasonable substitution because skipping is simply seeking at intervals, so pressing the fast forward button in a car and having the audio skip forward 15/30/60 seconds would be expected UX.  Also I did notice that there is no icons for seekForward and seekBackwards for the android notifications, though obviously those could be added later if desired.

Guichaguri:
@SteffeyDev you should use the fast forward and the fast rewind icons from [Material Icons](https://material.io/icons/)

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
SteffeyDev: Sorry, I am not an Android developer, what is the dex method limit?

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
SteffeyDev: I just changed this, check out my latest revision

Guichaguri:
Sorry, I meant PRs. Splitting into different PRs for each change should help a lot.

index.ios.js
SteffeyDev: Oh, you are right, but I think the better solution is to use a Platform switch in the same file instead of copying everything to a new file and just changing the constants.

SteffeyDev:
Just pushed a few updates regarding consolidation and skips, but I will break into separate PR's soon

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
Guichaguri: Android Java binaries work differently than regular java binaries, the methods are stored in a different way, the problem is that they have a limit of methods. It is possible to workaround it, but it's kinda annoying, mainly for someone that is using React Native and have no idea about how Android works.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
SteffeyDev: Are you suggesting that I copy your code over to work here?  I love the idea of react-native-track-player btw and will probably switch to that once it's ready for production.

SteffeyDev:
Closing in favor of #77, #78, and #79.  Those PR's should have addressed all of the changes addressed here except for detecting local images on Android, which we can continue discussing in #78.

PR #77, opened 2017-08-06
Add Skip control on Android

SteffeyDev:
#### What's this PR do?
This PR adds support for the skip control on android, by changing the image used by fast-forward and rewind to match the skip icon on iOS.

#### How to test:
Test by enabling the skipForward and skipBackward controls on Android, and testing to make sure that the buttons trigger the appropriate callbacks.

#### To Discuss
From #76, should the actual fast-forward and rewind icons be added as an option?  If this is included, then the app developer using this library would specify the control they want to use, and this would dictate the icon used in the Android notification.

For example, if they used ```MusicControl.enableControl('seekForward', true);``` and ```MusicControl.on('seekForward', () => {});``` then it would use the fast-forward icon, but if they used ```MusicControl.enableControl('skipForward', true);``` and ```MusicControl.on('skipForward', () => {});```, then it would use the skip icon.

I believe this would allow better cross platform support, because iOS currently supports both options.  That being said, I am open to discussion on the topic.


Guichaguri:
@tanguyantoine What is the difference between `seekForward` and `skipForward` in iOS?

SteffeyDev:
@Guichaguri while we wait for @tanguyantoine to respond, I can respond with a few thoughts.  According to https://developer.apple.com/documentation/mediaplayer/mpremotecommandcenter, the seeking is "seeking forward/backward through a single media item", while skipping is "playing a future/previous point in a media item."  

It is notable that on iOS 10, the seeking option does not actually place any buttons on the control panel, and is supposably only used for external devices that support seeking buttons.  Only the next/previous track and skip forward/backward add buttons to the control panel (obviously not including the play/pause/stop/etc central buttons).

Given that the "seek" option is already enabled for this library on Android only, it may be possible to just not allow "seekForward" and "seekBackward" on Android for this library, as I believe skipping is a more desired attribute and I'm not entirely sure what seeking *buttons* would actually do.  Seeking is usually associated with a scrubbing bar, which I believe is what you have available on Android by enabling the "seek" command.

tanguyantoine:
I answered but in the wrong issue 😬  https://github.com/tanguyantoine/react-native-music-control/pull/76#issuecomment-320456335

Guichaguri:
So instead of trying to support both `skipForward` and `seekForward` as they'd do the same thing on Android, why not just replace the `seekForward` to `skipForward`?

SteffeyDev:
Sorry it took me a while to get to this, @Guichaguri now seekForward is replaced with skipForward.  Let me know what you think.

Guichaguri:
The code looks good, but the icons still look like "replay" icons.

Use the "fast forward" and the "fast rewind" from [Material Icons](https://material.io/icons/) instead

SteffeyDev:
So I hate to belabor this, but [some research](https://en.wikipedia.org/wiki/Media_controls) indicates that the expected use of the fast forward and rewind buttons is to change the speed of the playback, not skip forward by a given interval.  The icons I used are reminiscent to the [iOS icons for skip](https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&cad=rja&uact=8&ved=0ahUKEwi5vP_X4v_VAhUG1CYKHY9MBMMQjRwIBw&url=https%3A%2F%2Fstackoverflow.com%2Fquestions%2F33915871%2Fswift-2-change-buttons-background-lock-screen-mediaplayer&psig=AFQjCNEiKZo8YOCMDwGYXkb1EXpK51v0mw&ust=1504210255784707) and that of other Android apps that use skip in their user interfaces.  However, I did miss one key part, which is that formally the skip button should have [2 arrows terminating the line and often have a number inside](https://channels-discourse.s3.amazonaws.com/original/2X/b/ba87dd021a21f7507c864a8343f3f224ead3bc1f.png), though I omitted the latter intentionally.

@Guichaguri If I change the icon to include the double arrows and possible internal tick marks, to differentiate it from the replay button, would that be sufficient, or will you only accept the fast-forward and re-wind icons?

Guichaguri:
@SteffeyDev Yes, that would be sufficient. You can use [Material Icons](https://material.io/icons/) "forward 10" and "replay 10" icons

SteffeyDev:
Oh, I was looking for those icons, thanks for pointing them out.  I can also add a switch in android to support the {interval: x} option that will choose the correct icon for the interval, and update the README to reflect this if you would like @Guichaguri.

Guichaguri:
I don't think bundling all the icons and change them based on the interval is a good idea, there would be a lot of useless resources in the app.
I think you should leave it as is, the icon can be overwritten by the app developer anyway.

SteffeyDev:
Just saw the last comment, and I went ahead and included all the icons before I saw it.  Anyways, it shouldn't matter because Material Icons are actually just .xml vector files, so they take up almost no space to store on device.  For older devices below API 21, the vectors are converted to PNGs during build, but they are also very small.  Let me know what you think of the changes in the last commit.

Guichaguri:
@SteffeyDev If we're going to use vector icons, why not do the same for all the other icons? They're all Material Icons anyway

SteffeyDev:
@Guichaguri I feel that changing the other icons is outside the scope of this PR, which is why I mentioned doing them in another PR.  If you would like to include in this PR regardless, let me know

Guichaguri:
@SteffeyDev Yeah, do it in another PR

Thanks :)

PR #78, opened 2017-08-06
Artwork Bug Fixes (Issue #46)

SteffeyDev:
#### What's this PR do?
This fixes a bug where a react native app running on the simulator
couldn’t load artwork, the iOS couldn’t load local artwork using the
require statement, and on Android the image was being cleared on every
updatePlayback call, as well as other improvements

Issue #46 

#### How to test:
Artwork needs to be tested locally and remotely on both platforms, and the ability to change out the artwork half way through playback needs to be tested as well.  In my app (which is currently using this code), I initially load dummy artwork and then if there is an internet connection, load the actual artwork after.  That way, if there is not an internet connection, there is still artwork with basic branding.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -187,16 +187,7 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
             rating = RatingCompat.newUnratedRating(ratingType);
         }
 
-        String artwork = null;
-        boolean localArtwork = false;
-        if(metadata.hasKey("artwork")) {
-            if(metadata.getType("artwork") == ReadableType.Map) {
-                artwork = metadata.getMap("artwork").getString("uri");
-                localArtwork = true;
-            } else {
-                artwork = metadata.getString("artwork");
-            }
-        }
+
 
         md.putText(MediaMetadataCompat.METADATA_KEY_TITLE, title);
         md.putText(MediaMetadataCompat.METADATA_KEY_ARTIST, artist);
@@ -212,7 +203,17 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         nb.setContentInfo(album);
         nb.setColor(notificationColor);
 
-        if(artwork != null) {
+        if(metadata.hasKey("artwork")) {
+            String artwork = null;
+            boolean localArtwork = false;
+
+            if(metadata.getType("artwork") == ReadableType.Map) {
+                artwork = metadata.getMap("artwork").getString("uri");
+                localArtwork = true;
+            } else {
+                artwork = metadata.getString("artwork");
+            }
+
             final String artworkUrl = artwork;
             final boolean artworkLocal = localArtwork;
 
@@ -361,7 +362,8 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         Bitmap bitmap = null;
 
         try {
-            if(local) {
+            // If we are running the app in debug mode, the "local" image will be served from htt://localhost:8080, so we need to check for this case and load those images from URL
+            if(local && !url.startsWith("http")) {
 
                 // Gets the drawable from the RN's helper for local resources
                 ResourceDrawableIdHelper helper = ResourceDrawableIdHelper.getInstance();
--- a/index.ios.js
+++ b/index.ios.js
@@ -6,6 +6,7 @@
 
 import { NativeModules, NativeEventEmitter } from 'react-native';
 const NativeMusicControl = NativeModules.MusicControlManager;
+import resolveAssetSource from 'react-native/Libraries/Image/resolveAssetSource';
 
 /**
  * High-level docs for the MusicControl iOS API can be written here.
@@ -40,7 +41,12 @@ var MusicControl = {
     NativeMusicControl.enableBackgroundMode(enable)
   },
   setNowPlaying: function(info){
-    NativeMusicControl.setNowPlaying(info)
+    // Check if we have an ios asset from react style image require
+    if(info.artwork) {
+        info.artwork = resolveAssetSource(info.artwork) || info.artwork;
+    }
+
+    NativeMusicControl.setNowPlaying(info);
   },
   resetNowPlaying: function(){
     NativeMusicControl.resetNowPlaying()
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -212,33 +203,47 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
+              md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, bitmap);
+              nb.setLargeIcon(bitmap);
+
+              session.setMetadata(md.build());
+              notification.show(nb, isPlaying);
+              artworkThread = null;
+          }
+      });
+      artworkThread.start();
+    }
```
Guichaguri: Again, why move to a new method if it's only called by `setNowPlaying`?

index.ios.js
```diff
@@ -40,7 +41,12 @@ var MusicControl = {
     NativeMusicControl.enableBackgroundMode(enable)
   },
   setNowPlaying: function(info){
-    NativeMusicControl.setNowPlaying(info)
+    // Check if we have an android asset from react style image require
```
Guichaguri: What do you mean by "android asset"?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
tanguyantoine: shorter method are more readable, no ?

ios/MusicControlManager.m
```diff
@@ -78,12 +78,17 @@ - (dispatch_queue_t)methodQueue
 
     center.nowPlayingInfo = [self update:mediaDict with:details andSetDefaults:false];
 
-    // Update the image if it exists
-    if ([details objectForKey:@"artwork"] != nil) {
+    // Custom handling of artwork in another thread, will be loaded async
```
tanguyantoine: It looks a lil bit redundant. any chance to have a dedicated method ?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
SteffeyDev: It's called in setNowPlaying and updatePlayback, so that you can update the artwork if you want to change it while the audio is currently playing.  I use this to load a generic artwork first and then load one over the network later.

index.ios.js
SteffeyDev: Oops, copy and paste strikes again, sorry.  This allows the loading of artwork from RN's require(), which was previously not supported on iOS.

ios/MusicControlManager.m
SteffeyDev: You are right, I can definitely break these checks into there own method.

SteffeyDev:
Latest revision (a310fb1) refactored the iOS code to move all of the image logic to one method, renamed to updateArtworkIfNeeded.  This is called in updatePlayback and setNowPlaying, similar to the Android code.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -212,33 +203,47 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
-                    md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, bitmap);
-                    nb.setLargeIcon(bitmap);
-
-                    session.setMetadata(md.build());
-                    notification.show(nb, isPlaying);
-                    artworkThread = null;
-                }
-            });
-            artworkThread.start();
+        if (metadata.hasKey("artwork")) {
```
Guichaguri: Can you keep the same code style throughout the code?
This `if` should not have a space after

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -212,33 +203,47 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
+          setArtwork(metadata);
         } else {
-            md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, null);
-            nb.setLargeIcon(null);
+          md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, null);
+          nb.setLargeIcon(null);
```
Guichaguri: Same problem here, code style. Keep the way it was, four spaces.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
Guichaguri: Why? `updatePlayback` should only be used for player related stuff, while `setNowPlaying` should be used for metadata, including artwork.

There is no point in duplicating the API in two different methods.

I suggest you to keep it the way it was, for two reasons: help prevent multidex issues and keep the API the way it was.
Also, the code style here is different from the rest of the code too.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
SteffeyDev: Ah, didn't catch that coding style attribute, thanks

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
SteffeyDev: Ok, that clarifies things.  For some reason I was under the impression that setNowPlaying was used for the first time to set everything up and updatePlayback was used every time after that.  I'll change it back.

SteffeyDev:
I made the changes that @Guichaguri requested.  The only reservations I have toward these requested changes is that if I want to update just the artwork, I now have to call setNowPlaying again, and if I don't pass in the original values for all of the fields that I want to use then it will wipe then.  @Guichaguri What is the best way to update just the artwork midway through the audio playing without touching the other properties?

Guichaguri:
@SteffeyDev This is the only way, unfortunately. You can update parts of it, but the changes will be pushed into a whole new object anyway. It's not a pricey update, just keep the object you've been passing to `setNowPlaying` and change parts of it.

Guichaguri:
Looks good. Thanks :)

SteffeyDev:
@Guichaguri Ok, you're right that shouldn't be too difficult, thanks for the help merging this!

PR #79, opened 2017-08-06
Consolidating Files

SteffeyDev:
#### What's this PR do?
Because the main file can be specified in package.json, it is not
necessary to have both index.android.js and index.ios.js.  This PR merges the two into one and uses Platform.OS where necessary to specify per-platform deviations.

This change will make it easier to add features to the javascript moving forwards, because you won't have to make the same changes in two different files.

#### How to test:
This PR should not affect the functionality of the code.


Guichaguri:
Merging the two files can make it harder to solve platform specific issues, but it does reduce duplicated code. I don't know if it's a good trade off.

The React Native packager would remove index.android in iOS and index.ios in Android anyway.

tanguyantoine:
let' s keep this one on hold in this case

tanguyantoine:
closed in favor of #120

PR #80, opened 2017-08-07
fix: only update image if it doesn't exist

ndillon1:
#### What's this PR do?
Only set a new artwork url if it didn't previously exist - iOS fix only 

updatePlayback function was updating the image every time it was called which was causing memory leaks

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -77,13 +77,13 @@ RCT_EXPORT_METHOD(updatePlayback:(NSDictionary *) originalDetails)
     NSMutableDictionary *mediaDict = [[NSMutableDictionary alloc] initWithDictionary: center.nowPlayingInfo];
 
     center.nowPlayingInfo = [self update:mediaDict with:details andSetDefaults:false];
-
-    // Update the image if it doesn't exists
-    if ([details objectForKey:@"artwork"] == nil) {
+    
+    // Update the image if only it it differs from the previous image
+    
+    if ([details objectForKey:@"artwork"] != self.artworkUrl) {
         self.artworkUrl = details[@"artwork"];
+        [self updateNowPlayingArtwork];
     }
-      [self updateNowPlayingArtwork];
-
 
 }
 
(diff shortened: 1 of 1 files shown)


ios/MusicControlManager.m
```diff
@@ -78,12 +78,13 @@ - (dispatch_queue_t)methodQueue
 
     center.nowPlayingInfo = [self update:mediaDict with:details andSetDefaults:false];
 
-    // Update the image if it exists
-    if ([details objectForKey:@"artwork"] != nil) {
+    // Update the image if it doesn't exists
+    if ([details objectForKey:@"artwork"] == nil) {
```
tanguyantoine: should we do that instead ? 
```obj-c
if ([details objectForKey:@"artwork"] != self.artworkUrl) {
  [self updateNowPlayingArtwork];
}
```

tanguyantoine:
@ndillon1 see my comment. Good catch 👍

ios/MusicControlManager.m
ndillon1: yeh that works too, I can update my PR with those changes

ndillon1: @tanguyantoine updated PR with those changes

tanguyantoine:
It looks like you branch is not up to date. could create a new PR with a new branch up to date ?

Jeremaiha:
Could someone here please help me out with instantiating the image url with require?
I tried the following code:
`loadCallToMusicControl() {
    MusicControl.setNowPlaying({
      title: this.title,
      artwork: require('../Images/tempoImage.png')`
But the app crashes.

tanguyantoine:
logs ?

Jeremaiha:
I managed to understand the problem.
First of all it needs to receive a `filePath` and not `RN's require()`, since it doesn't know how to handle it, it checks the prefix if it contains `http\s` so obviously it crashes for `require()`.
After that it checks if a file exist in this path, and here is where I fail to understand why it doesn't find my image in the correct path.

PR #81, opened 2017-08-11
fix: only update image if it the url has changed

ndillon1:
#### What's this PR do?

Fixes the issue where the updatePlayback function requests a new image every time which causes memory leaks


ndillon1:
Closing as it seems my PR was already merged

PR #84, opened 2017-08-28
Multiple options for Swipe-to-close on Android Notification

SteffeyDev:
#### What's this PR do?

Now the app developer can specify whether the Android audio notification will always
close on swipe, only close if the audio is paused, or will never close
when swiped.  In addition, a new control was added to provide a
callback for when it is closed via swipe, so that the app developer can
choose to terminate the audio or do nothing.

None, it is a convenience feature that other developers may find useful.

#### How to test:

Run on device, enable the closeNotification control, optionally pass in a 'when' qualifier option, and try closing the notification.  If it is allowed to be closed, it should call the enabled callback on notification close, so that the app can react appropriately.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -54,6 +54,8 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     private long controls = 0;
     protected int ratingType = RatingCompat.RATING_PERCENTAGE;
 
+    public NotificationClose notificationClose = NotificationClose.PAUSED;
+
     public MusicControlModule(ReactApplicationContext context) {
         super(context);
     }
@@ -109,14 +111,14 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
 
         state = pb.build();
 
-        notification = new MusicControlNotification(context);
+        notification = new MusicControlNotification(this, context);
         notification.updateActions(controls, null);
 
         IntentFilter filter = new IntentFilter();
         filter.addAction(MusicControlNotification.REMOVE_NOTIFICATION);
         filter.addAction(MusicControlNotification.MEDIA_BUTTON);
         filter.addAction(Intent.ACTION_MEDIA_BUTTON);
-        receiver = new MusicControlReceiver(this, context.getPackageName());
+        receiver = new MusicControlReceiver(this, context);
         context.registerReceiver(receiver, filter);
 
         context.startService(new Intent(context, MusicControlNotification.NotificationService.class));
@@ -351,6 +353,17 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
                     session.setPlaybackToLocal(AudioManager.STREAM_MUSIC);
                 }
                 return;
+            case "closeNotification":
+                if(enable) {
+                    if(options.getString("when").equals("always")) {
+                        this.notificationClose = notificationClose.ALWAYS;
+                    } else if(options.getString("when").equals("paused")) {
+                        this.notificationClose = notificationClose.PAUSED;
+                    } else {
+                        this.notificationClose = notificationClose.NEVER;
+                    }
+                    return;
+                }
             default:
                 // Unknown control type, let's just ignore it
                 return;
@@ -430,5 +443,11 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     public void onLowMemory() {
         Log.w("MusicControl", "Control resources are being removed due to system's low memory (Level: MEMORY_COMPLETE)");
         destroy();
+    }
+
+    public enum NotificationClose {
+        ALWAYS,
+        PAUSED,
+        NEVER
     }
 }
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
@@ -21,12 +21,14 @@ public class MusicControlNotification {
     protected static final String PACKAGE_NAME = "music_control_package_name";
 
     private final ReactApplicationContext context;
+    private final MusicControlModule module;
 
     private int smallIcon;
     private NotificationCompat.Action play, pause, stop, next, previous, skipForward, skipBackward;
 
-    public MusicControlNotification(ReactApplicationContext context) {
+    public MusicControlNotification(MusicControlModule module, ReactApplicationContext context) {
         this.context = context;
+        this.module = module;
 
         Resources r = context.getResources();
         String packageName = context.getPackageName();
@@ -67,8 +69,15 @@ public class MusicControlNotification {
         if(next != null) builder.addAction(next);
         if(skipForward != null) builder.addAction(skipForward);
 
-        // Notifications of playing music can't be removed
-        builder.setOngoing(isPlaying);
+        // Set whether notification can be closed based on closeNotification control (default PAUSED)
+        if(module.notificationClose == MusicControlModule.NotificationClose.ALWAYS) {
+            builder.setOngoing(false);
+        } else if(module.notificationClose == MusicControlModule.NotificationClose.PAUSED) {
+            builder.setOngoing(isPlaying);
+        } else { // NotificationClose.NEVER
+            builder.setOngoing(true); 
+        }
+
         builder.setSmallIcon(smallIcon);
 
         // Open the app when the notification is clicked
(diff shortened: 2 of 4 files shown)


Guichaguri:
I can't think of one scenario where someone would want to change the swipe controls.

In the Android documentation, there's a music player article where they recommend making the notification "on going" when the music is playing, so the user will always have some way to control it and be prevented to remove it accidentally.

SteffeyDev:
Ok here's the story: I was approached by one of my Android beta testers with a concern: if he swiped the notification closed while the audio was paused, he didn't know how to get it back. Of course, when trying to demo this for me, he wasn't able to swipe it closed at all, because at the time neither of us knew that currently, you can only swipe the notification closed while the audio is paused.  I actually had to dig into the Android code to find why this is so.

Anyways, he said that he expected that when he swiped the notification close, he expected the audio to stop playing, but it was frustrating because he couldn't always swipe it closed and this confused him. As an iOS user myself, I didn't have any ground in which to tell him that his expectations were off base, and agreed that it would make sense for you to always be able to swipe the notification closed. 

As a developer, this revealed another need: to get a callback when the notification was swiped closed so that I could update the US accordingly when they came back to the app next. Currently, if they paused the audio and close the notification, my app had no way of knowing of this. 

Hence, this PR was born. I can imagine situations in which some apps may not want the user to be able to dismiss the audio by swiping the notification closed, and I can imagine apps like mine where swiping the notification closed at any point makes sense to end the audio. I can't as easily imagine a situation in which you would only want to swipe the notification closed when the audio is paused, but because that is the way it currently is, I kept that as the default for backward compatibility.

SteffeyDev:
So after a quick poll of Android music apps (like I said I am an iOS user), it appears that they all also support swipe to close on pause, and no swipe when playing audio.  A few of them have a physical close button in addition to being able to swipe to close on pause.  This seems very un-intuitive, because in the case of the close button, there are two ways to close the notification, and in the case of only being able to swipe when paused, this caused confusion for me and my users because is is not naturally clear.

At the very least, this PR can allow a callback to be established for when the audio is paused and then swiped to close, so that I can update my UI accordingly.  However, I do see value in the ability to always swipe, because I am playing audio that is approaching an hour in length and some people may just want to swipe it away without having to pause it first.

Guichaguri:
> So after a quick poll of Android music apps (like I said I am an iOS user), it appears that they all also support swipe to close on pause, and no swipe when playing audio.

Yes, that is the expected behavior on all Android media apps. Why? Because audio **has** to be played in a service, and for the service to be kept alive, it needs to be set as a "foreground service".
Foreground services require a on-going notification for various reasons (e.g. let the user know that something is using a lot of processing power and battery in background), that's why they can't be swiped.

React Native apps do it the "wrong way". They add the audio code in the UI part, which can cause a lot of performance and UX problems.
I've been working on a new module called [react-native-track-player](https://github.com/Guichaguri/react-native-track-player). It is still in development, but it aims to provide audio playback, music controls and chromecast support the way it should've been: in a service, everything all tied together.

> A few of them have a physical close button in addition to being able to swipe to close on pause. This seems very un-intuitive, because in the case of the close button, there are two ways to close the notification, and in the case of only being able to swipe when paused, this caused confusion for me and my users because is is not naturally clear.

That is because there was a bug in pre-lollipop devices that caused notifications that were added with foreground services never remove the "on-going" property, even after the foreground mode was disabled.
To fix that bug, apps can provide a "close" button. Some of them provide the close button in all Android versions, [others](https://github.com/Guichaguri/react-native-track-player/blob/dev/android/src/main/java/guichaguri/trackplayer/metadata/components/MediaNotification.java#L118-L129) provide it only in pre-lollipop devices.

SteffeyDev:
Ok, so react-native-track-player aside, I think you have explained why the "notificationClosed" callback is completely necessary, so that when it is swiped closed after pause the react native app can clean up the audio resources used by whatever sound library they are using.  I would be fine dropping the {when: x} option and just allowing users to register for the "notificationClosed" callback  on the RN side.

Guichaguri:
The problem is definitely not the "notificationClosed" callback, but the option to change the on-going property, as it changes the expected behavior for Android music apps.

Anyway, this commit only works because this module is not tied together with the music player in a service, so I suppose I shouldn't be filtering new features because of their UX behavior, just tell me when you're done so I can accept it

SteffeyDev:
Ok, I just fixed a typo and it is ready to be pushed now @Guichaguri.  If it makes you feel any better, the default behavior is still the expected behavior, and developers can completely ignore the enableControl part of this and just give it a callback if they want the default behavior.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -293,9 +300,20 @@ synchronized public void resetNowPlaying() {
     }
 
     @ReactMethod
-    synchronized public void enableControl(String control, boolean enable) {
+    synchronized public void enableControl(String control, boolean enable, ReadableMap options) {
         init();
 
+        if (control.equals("closeNotification") && enable) {
```
Guichaguri: Why not put this inside the `switch`?

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -50,21 +52,22 @@ public void show(NotificationCompat.Builder builder, boolean isPlaying) {
         if(stop != null) builder.addAction(stop);
         if(next != null) builder.addAction(next);
 
-        // Notifications of playing music can't be removed
-        builder.setOngoing(isPlaying);
+        // Set whether notification can be closed based on closeNotification control (default PAUSED)
+        if (this.module.notificationClose == MusicControlModule.NotificationClose.ALWAYS) builder.setOngoing(false);
```
Guichaguri: Use the same code style here by adding the brackets

Also, remove `this` to make it cleaner.

android/src/main/java/com/tanguyantoine/react/MusicControlReceiver.java
```diff
@@ -28,6 +35,11 @@ public void onReceive(Context context, Intent intent) {
             module.notification.hide();
             module.session.setActive(false);
 
+            // Notify react native
+            WritableMap data = Arguments.createMap();
+            data.putString("name", "closeNotification");
+            this.reactContext.getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class).emit("RNMusicControlEvent", data);
```
Guichaguri: Again, there is no need to use `this` here

index.android.js
```diff
@@ -47,8 +47,8 @@ var MusicControl = {
   resetNowPlaying: function(){
     NativeMusicControl.resetNowPlaying()
   },
-  enableControl: function(controlName, enable){
-    NativeMusicControl.enableControl(controlName, enable)
+  enableControl: function(controlName, enable, options = {}){
+    NativeMusicControl.enableControl(controlName, enable, options || {})
```
Guichaguri: Wouldn't adding `options = {}` and then `options || {}` be redundant?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -54,6 +54,14 @@
     private long controls = 0;
     protected int ratingType = RatingCompat.RATING_PERCENTAGE;
 
+    public enum NotificationClose {
```
Guichaguri: Can you move the enum to the end of the class? Makes it more organized

index.android.js
SteffeyDev: I'm copying what was already in index.ios.js.  I thought it was redundant as well but it also seemed weird to have a different syntax for the exact same thing in both files.

SteffeyDev:
@Guichaguri changes made and ready for review.

tanguyantoine:
@Guichaguri do you have a npm account ? will grqnt you rights to publish :)

Guichaguri:
@tanguyantoine Yes, my username is `guichaguri`

android/src/main/java/com/tanguyantoine/react/MusicControlReceiver.java
```diff
@@ -4,15 +4,22 @@
 import android.content.Context;
 import android.content.Intent;
 import android.view.KeyEvent;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
+
```
Guichaguri: Unnecessary line break

tanguyantoine:
done

SteffeyDev:
@Guichaguri Ok, I fixed the tabs to spaces and removed the extra line.  Once you finish requesting changes, let me know and I will squash the commits and rebase on master to resolve the conflicts.

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -50,21 +52,26 @@ public void show(NotificationCompat.Builder builder, boolean isPlaying) {
...
-        if(!isPlaying) {
-            // Remove notification
-            Intent remove = new Intent(REMOVE_NOTIFICATION);
-            remove.putExtra(PACKAGE_NAME, context.getApplicationInfo().packageName);
-            builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, remove, PendingIntent.FLAG_UPDATE_CURRENT));
-        }
+        // Remove notification
+        Intent remove = new Intent(REMOVE_NOTIFICATION);
+        remove.putExtra(PACKAGE_NAME, context.getApplicationInfo().packageName);
+        builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, remove, PendingIntent.FLAG_UPDATE_CURRENT));
```
Guichaguri: If this intent is not conditional anymore, you should move it to the constructor

SteffeyDev: I could do that, but the builder object is local to this method, so I would have to store the intent off as a class variable.  If I were to do that, it would make sense to do the same to the openApp intent, which was not conditional originally but is still constructed in this method.

Guichaguri:
Fix the conflicts and then I'll be able to merge it :)

SteffeyDev:
@Guichaguri ok, ready to merge

Guichaguri:
Thanks

PR #85, opened 2017-09-01
Converting remaining icons to XML vector files

SteffeyDev:
#### What's this PR do?

Convert all the remaining icons to XML files, which saves space on newer devices and allows for density-independent image rendering (read more [here](https://android-developers.googleblog.com/2015/09/android-studio-14.html)).

#### How to test:

Test normal usage to confirm that all icons appear as normal.  Also need test on older devices (lower than api 21) to make sure that the pngs are generated and loaded correctly for backward compatibility.


Guichaguri:
Looks good. Thanks :)

PR #88, opened 2017-09-10
Add custom notification icon options for Android

mzmiric5:
#### What's this PR do?
This feature allows library users to supply their own custom icons to be used by the notification on Android. This was technically already possible, but the custom icon lookup string used previously required the asset to contain `-` which is not supported by Android's resource system.

The PR also adds a `setCustomNotificationIcon` function that allows the developer to use a custom resource name.

Not related to an issue, but to a feature I needed.

#### How to test:
To test the PR, you have to create a drawable resource in your Android app. In the resource, you should add an icon named `music_control_icon.png`. You can just add it top level to drawable or use multiple resolutions created by various tools or Android Studio's notification icon UI. Now the notification should be shown with a custom icon.

Now add another icon with a different name, for example `my_custom_icon.png`.
In your code add
```javascript
  MusicControl.setCustomNotificationIcon('my_custom_icon');
```
Now the notification should be shown with a custom icon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,17 @@ componentDidMount() {
 }
 ```
 
+### Customization
 
+It is possible to customize the icon used in the notification on Android.
+By default you can add a drawable resource to your package with the file name `music_control_icon` and the notification will use your custom icon. If you
+would like to use a custom icon name, you can use the `setCustomNotificationIcon`
+function and pass in your drawable resource name in as a string. Keep in mind
+that the resource must be in the drawable sub-package in your app.
+
+```javascript
+  MusicControl.setCustomNotificationIcon('my_custom_icon');
+```
 
 # TODOS
 
--- a/README.md
+++ b/README.md
@@ -115,7 +115,8 @@ MusicControl.setNowPlaying({
   description: '', // Android Only
   color: 0xFFFFFF, // Notification Color - Android Only
   date: '1983-01-02T00:00:00Z', // Release Date (RFC 3339) - Android Only
-  rating: 84 // Android Only (Boolean or Number depending on the type)
+  rating: 84, // Android Only (Boolean or Number depending on the type)
+  notificationIcon: 'my_custom_icon' // Android Only (String), Android Drawable resource name for a custom notification icon
 })
 ```
 
@@ -269,10 +270,10 @@ componentDidMount() {
 ### Customization
 
 It is possible to customize the icon used in the notification on Android.
-By default you can add a drawable resource to your package with the file name `music_control_icon` and the notification will use your custom icon. If you
-would like to use a custom icon name, you can use the `setCustomNotificationIcon`
-function and pass in your drawable resource name in as a string. Keep in mind
-that the resource must be in the drawable sub-package in your app.
+By default you can add a drawable resource to your package with the file name `music_control_icon` and the notification will use your custom icon.
+If you need to specify a custom icon name, or change your notification icon during runtime, the `setNowPlaying` function accepts a string 
+for an Android drawable resource name in the `notificationIcon` prop. Keep in mind that just like with `music_control_icon` the resource specified has
+to be in the drawable package of your Android app.
 
 ```javascript
   MusicControl.setCustomNotificationIcon('my_custom_icon');
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -161,6 +161,12 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     }
 
     @ReactMethod
+    public void setCustomNotificationIcon(String resourceName) {
+        init();
+        notification.setCustomNotificationIcon(resourceName);
+    }
+
+    @ReactMethod
     synchronized public void setNowPlaying(ReadableMap metadata) {
         init();
         if(artworkThread != null && artworkThread.isAlive()) artworkThread.interrupt();
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -54,8 +54,6 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     private boolean isPlaying = false;
     private long controls = 0;
     protected int ratingType = RatingCompat.RATING_PERCENTAGE;
-
-    private boolean hasCustomIcon = false;
     
     public NotificationClose notificationClose = NotificationClose.PAUSED;
 
@@ -128,7 +126,6 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
 
         context.registerComponentCallbacks(this);
 
-        hasCustomIcon = false;
         isPlaying = false;
         init = true;
     }
@@ -156,7 +153,6 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         pb = null;
         nb = null;
 
-        hasCustomIcon = false;
         init = false;
     }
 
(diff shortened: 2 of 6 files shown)


android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -34,10 +34,18 @@ public MusicControlNotification(MusicControlModule module, ReactApplicationConte
         String packageName = context.getPackageName();
 
         // Optional custom icon with fallback to the play icon
-        smallIcon = r.getIdentifier("music-control-icon", "drawable", packageName);
+        smallIcon = r.getIdentifier("music_control_icon", "drawable", packageName);
```
mzmiric5: this change was necessary, as Android build would fail if a drawable with resource name containing a `-` was created. Using `_` resolves the issue.

README.md
```diff
@@ -266,7 +266,17 @@ componentDidMount() {
 }
 ```
 
+### Customization
 
+It is possible to customize the icon used in the notification on Android.
+By default you can add a drawable resource to your package with the file name `music_control_icon` and the notification will use your custom icon. If you
```
Guichaguri: As you already know, this is an option. Why add a new function for that? It would only add complexity and redundancy to this module.

Also, if I were you, I would accept a `require` resource instead of just using a string

mzmiric5: Thanks for the review. I guess I really didn't make a good case for the function part of this PR, so let me try to do that here. One of the features I was looking for was the ability to specify the notification icon I want to use on the fly. Say my app plays various types of media, and right now it plays an audio book. I would like it to show a book icon in the notification. Then the user switches to listening to a podcast, and I would like to show a microphone icon in the notification instead. With just the custom option, this would not be possible, but having the function allows me to basically point the notification to a different resource on the fly, and I one of the ways I use this feature. I believe that justifies it being a function, but if you still believe it's not necessary, I can remove it from the PR.

Regarding using a string, I went the drawable resource route, because Android resources basically provide you with a free multi resolution support. So if you use PNGs, Android drawable resources allow you to generate files of resolutions that match various device resolutions. I think having a `require` resource is a good idea, but I think it should be in a separate function because it provides very different functionality and adds more complexity while losing out on built in multi resolution support. Let me know what you think about this.

Guichaguri: Thanks for clarifying. Is there any reason to create a new function? Can't this be an option in `setNowPlaying` instead?

mzmiric5: ok, the latest update to the PR should take care of this. let me know if there are any other issues

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -205,6 +211,16 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
+        hasCustomIcon = notificationIcon != null ? true : false;
+        if (hasCustomIcon) {
+          int smallIcon = loadIconResource(notificationIcon);
+          if (smallIcon != 0) {
+            nb.setSmallIcon(smallIcon);
+          } else {
+            hasCustomIcon = false;
+          }
+        }
+
```
Guichaguri: Can't you bring back `setCustomNotificationIcon` from the notification class and just call it from here? It would reduce a lot of code. Also, there is no reason for `hasCustomIcon`

mzmiric5: so basically, the setCustomNotificationIcon was replacing the value of `smallIcon` class variable in the Notification. If it was used with the `setNowPlaying`, it would mean that if the developer wanted to change the icon for a specific nowPlaying event, but then didn't specify it on the next call of `setNowPlaying` they might expect a default icon they had, but now that would be overwritten by the icon they specified in the previous call. Doing it this way, allows us to set the notification on the individual call to `setNowPlaying` if specified as an option, and fall back to default if not specified.

The `hasCustomIcon` was then added to prevent any calls from show to run `builder.setSmallIcon` inside of the notification code, as that would overwrite the custom icon. It could have been used as a local variable in the `setNowPlaying` and just passed down, but since `updateNowPlaying` also makes a conditional call to `show` that could cause the custom icon to be overwritten.

Guichaguri: If no custom icon is specified, you can load the `play` icon back

mzmiric5: Basically, here's my train of thought, and I might be wrong.

Developer has an icon `music_control_icon` that they use as the default one for their notifications. When they want to use a custom icon, they specify `notificationIcon` to something like `my_custom_icon` in the `setNowPlaying` in some specific case. Then they call `setNowPlaying` again without the the `notificationIcon`, they expect the notification icon to be `music_control_icon`

In the original code I wrote, the `setCustomNotificationIcon` would set the value of `smallIcon` variable of the `MusicControlNotification`. If `setNowPlaying` called `setCustomNotificationIcon` when a `notificationIcon` was set, it would basically overwrite the `smallIcon` variable value in the instance of the notification, which will result in the icon specified once being used for any future notifications, even if `notificationIcon` was not specified.

The version of code I pushed in the latest update of this PR does the following. `smallIcon` from the `MusicControlNotification` never gets touched, so after the instance is created, the value of `smallIcon` will be a reference either to `play` or to `music_control_icon` if it was provided. Instead what it does, is if the developer passed down a `notificationIcon` to the `setNowPlaying` call, it would try to load that resource, and if the resource is available it would set it on the NotificationBuilder instance being passed down to the `show` call. The obvious result of this is if the `notificationIcon` is not provided, the NotificationBuilder being passed down to the show call would not have an icon set on it, and the notification uses whatever the value of `smallIcon` was set to when the Notification instance was originally created.

Which then brings us back to the `hasCustomIcon` boolean flag, and the need for it. In the `MusicControlNotification` the `show` method will call `builder.setSmallIcon(smallIcon)`. Because the change in `setNowPlaying` and it being a safer way to not touch the value of `smallIcon`, any call to the `show` function would result in the icon being set to whatever the value of `smallIcon` is, thus overriding the change from the `setNowPlaying` call. However if we can pass something like `hasCustomIcon` we can basically prevent that from happening.

Maybe I'm missing something, but regarding the code reduction, I don't see how it would be more or less code, or in any way affect the complexity? It's still essentially the same code.

Guichaguri: The way I would do it is:
* Create a `customIcon` field in `MusicControlNotification`
* Bring back `setCustomNotificationIcon`, but setting `customIcon` instead of `smallIcon`
* Accept null values in `setCustomNotificationIcon`, which would set `customIcon` to -1
* Check whether `customIcon` is -1 in `MusicControlNotification.show`, if it isn't, then use it instead of `smallIcon`

There is also an even simpler way to achieve this:
* Bring back `setCustomNotificationIcon`, but using `builder.setSmallIcon` directly
* Accept null values in `setCustomNotificationIcon`, which would `builder.setSmallIcon` with the `smallIcon` field
* Remove calls to `builder.setSmallIcon` in `MusicControlNotification.show`

That way, it would be much cleaner. Is there any other reason to use your code that I'm not seeing? Thanks :)

mzmiric5: ok, cool, wasn't sure we wanted to rip out `setSmallIcon` from `show`. With that in mind, yeah let me just push an update, that should make this simpler. Thanks again

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -34,9 +35,16 @@ public MusicControlNotification(MusicControlModule module, ReactApplicationConte
...
         if(smallIcon == 0) smallIcon = r.getIdentifier("play", "drawable", packageName);
     }
+    
+    public void setCustomNotificationIcon(String resourceName) {
+      Resources r = this.context.getResources();
+      String packageName = this.context.getPackageName();
+
+      customIcon = r.getIdentifier(resourceName, "drawable", packageName);
```
Guichaguri: This is the last thing: you're not following the same code style, using two spaces instead of four.
Also, you can remove `this` from the context, it's not really needed

Other than that, it looks good :+1:

Guichaguri:
Thank you :)

tanguyantoine:
@Guichaguri I published the new version 0.4.9

Issue #101, opened 2017-10-17
Headphones unplugged trigger for Android


jankarres:
# Description

The package is able to trigger "pause" on headphones unplugged for iOS. This would be a nice feature for Android, too. I this feature should not be complicated to implement think for Android developer in Java. Actually you could help out, @tanguyantoine? Thanks so much for the great work!

Example code: https://twigstechtips.blogspot.de/2013/06/android-detect-when-headsetheadphones.html

Platform
   - [ ] iOS
   - [x] Android
   
Device
  - [ ] Simulator
  - [X] Real device

tanguyantoine:
@Guichaguri any idea ?

Guichaguri:
Yeah, pretty easy to implement, I can work on it when I have time

frankros91:
I don't normally code in Android, but I took a try at this yesterday and it seems to be working.  Might not be the ideal way to do this.  But set up the listener to fire the same PAUSE KeyEvent that is fired when the pause button is pressed in the notification try.

PR #104, opened 2017-11-01
implemented pause functionality when the headphones are unplugged for Android

frankros91:
#### What's this PR do?
Android Only

Implements logic in MusicControlReceiver.java that listens to when the headphones are unplugged.  If unplugged, a pause key event is dispatched, similar to if the pause button was pressed in the notification tray.
Issue 101
https://github.com/tanguyantoine/react-native-music-control/issues/101


Guichaguri:
1. You should use the same broadcast receiver, no need to use another one
2. Don't directly use strings for intent actions, use the constant instead
3. Fix the code style, reformat the code

jankarres:
PR refers to issue #101

Guichaguri:
If you want, I can work on this tomorrow

frankros91:
Not sure why the indentations are all off, might be my editor.  I'll clean up and make a final push in a bit

jankarres:
@Guichaguri Do you see any chance to review this PR in the next few days? This would be awesome!

Guichaguri:
There are still code style issues and the code can be simplified with `AudioManager.ACTION_AUDIO_BECOMING_NOISY`. I'll probably push a commit later today to fix it

jankarres:
Amazing. Thanks @Guichaguri and @frankros91!

PR #108, opened 2017-11-15
skip backward default button default erroneously left out a 10 in the…

DanNi0130:
The default SkipBackward icon left off the number of seconds so no icon appears by default on android. This PR fixes the issue by setting the default icon to 10 seconds (like it is for SkipForward).

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
@@ -67,7 +67,7 @@ public class MusicControlNotification {
         if (options != null && options.containsKey("skipBackward") && (options.get("skipBackward") == 10 || options.get("skipBackward") == 5 || options.get("skipBackward") == 30)) {
             skipBackward = createAction("skip_backward_" + options.get("skipBackward").toString(), "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
         } else {
-            skipBackward = createAction("skip_backward_", "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
+            skipBackward = createAction("skip_backward_10", "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
         }
     }
 


android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -67,7 +67,7 @@ public synchronized void updateActions(long mask, Map<String, Integer> options)
         if (options != null && options.containsKey("skipBackward") && (options.get("skipBackward") == 10 || options.get("skipBackward") == 5 || options.get("skipBackward") == 30)) {
             skipBackward = createAction("skip_backward_" + options.get("skipBackward").toString(), "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
         } else {
-            skipBackward = createAction("skip_backward_", "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
+            skipBackward = createAction("skip_backward_10", "Skip Backward", mask, PlaybackStateCompat.ACTION_REWIND, skipBackward);
```
Guichaguri: That's actually supposed to be there. There [isn't a `skip_backward_` icon](https://github.com/tanguyantoine/react-native-music-control/tree/master/android/src/main/res/drawable).

I don't think I get what is the issue...

DanNi0130: Yeah, the problem is that there is no icon that corresponds to skip_backward_, there are icons for skip_backward_5, skip_backward_10, and skip_backward_30. The new code is just setting the default icon to skip_backward_10 (the default icon for skipforward is skip_forward_10).

DanNi0130: I can create a few gifs later today to show what this code is doing.

Guichaguri: Oh, I don't know what I was thinking, I thought you removed the `10` instead of adding it :P

Thank you for the PR :)

DanNi0130: No worries :).

tanguyantoine:
I published version 0.4.13

Issue #98, opened 2017-09-27
Artwork from url not showing


natashache:
# Description
Music control is not showing the album artwork from url.

![img_4128](https://user-images.githubusercontent.com/19754530/30918040-64d52382-a36b-11e7-8566-3256e421d93d.PNG)
1. Sample code 
```
            MusicControl.setNowPlaying({
                title: `${episode.title}`,
                artwork: 'https://s3.amazonaws.com/soundwiseinc/demo/d4c68be0-9d6c-11e7-81ba-314068aa6d3b.png',
                artist: _currentPublisher.name,
                album: _currentSoundcast.title,
                duration: Playing._duration, // (Seconds)
                description: episode.description, // Android Only
                color: 0xFFFFFF, // Notification Color - Android Only
                date: moment(episode.date_created * 1000).utc().format(), // Release Date (RFC 3339) - Android Only
                rating: episode.likes && Object.keys(episode.likes).length || 0 // Android Only (Boolean or Number depending on the type) //
            });
```

2. Platform ?
   - [x ] iOS
   
3. Device
  - [x ] Real device

3. Modules
```
    "react": "16.0.0-alpha.12",
    "react-native": "0.46.4",
    "react-native-music-control": "git+https://github.com/tanguyantoine/react-native-music-control.git",
```

martinlockett:
Works for me on iOS10 (looks like you're using iOS10)

```
    "react": "16.0.0-alpha.6",
    "react-native": "0.44.2",
```

tanguyantoine:
do you have a source code we look at?

egorkhmelev:
I have experienced the same issue and looks like I have found the cause.

1. I call `MusicControl.setNowPlaying` with `artwork` url in it. It gonna reset artwork here https://github.com/tanguyantoine/react-native-music-control/blob/master/ios/MusicControlManager.m#L96 and will try to load remote image in background to update artwork to later.
2. Then I will call `MusicControl.updatePlayback` from time to time without artwork at all (Statement from docs: You don't need to call this method filling all properties) and it will reset `this.artworkUrl` to nil here https://github.com/tanguyantoine/react-native-music-control/blob/master/ios/MusicControlManager.m#L83
3. Remote image done loading and component will try to set it but fail here https://github.com/tanguyantoine/react-native-music-control/blob/master/ios/MusicControlManager.m#L281 as `this.artworkUrl` is already nil (from 2).

[...]

PR #109, opened 2017-11-15
Fix artwork url overwrite on ios

DanNi0130:
Fixes the issue in #98, adds a check to make sure artwork is not set to nil when updatePlayback is called. Tested on iPhone 6s.


DanNi0130:
Accidentally lumped two PRs together, I'll submit this one again separately.

Issue #107, opened 2017-11-10
[ios] Seek to position


kesha-antonov:
Hello guys!

Is it possible seek by tapping and moving progress bar of Music Control on iOS?

If it is not implemented yet can you give some advices? I'll provide PR later

Thanks

1. Platform ?
   - [x] iOS
   - [ ] Android
   
@tanguyantoine

Ayc0:
It's not working on my phone either

tanguyantoine:
I'm not sure to understand what you are expecting. Can you elaborate?

kesha-antonov:
On iPhone on lock screen you can tap-and-hold progress bar and move to right/left to desired position in seconds. It's not responding now

@tanguyantoine

tanguyantoine:
Should be fixed. I’ll push a new version soon

kesha-antonov:
Awesome! Thanks! @tanguyantoine

tanguyantoine:
thanks to @DanNi0130

PR #110, opened 2017-11-15
allow changing playback position by dragging slider on ios lock screen

DanNi0130:
This PR resolves issue #107, it allows you to change the playback position by dragging the slider on the iOS lock screen. Here is a gif of the functionality: 
demo change playback position

Tested on a real device (iPhone 6s) and the iPhone 7 simulator.


tanguyantoine:
I published 0.4.13

Issue #114, opened 2017-12-06
Add a stopControl function


Ayc0:
# This is a request a *feature*

Currently `MusicControl.resetNowPlaying()` is just disable the display of the control panel, but you can still interact with it. 

Is it possible to add a `MusicControl.stopControl()`function that calls `MusicControl.resetNowPlaying()` and remove every controls and if it is possible, that displays the previous control panel (for instance, you were listening some music on Spotify, when `stopControl` is called, the panel of Spotify is redisplayed)

PR #106, opened 2017-11-10
Fix set artwork. Remove warning for RN 0.50. Fix set listener

kesha-antonov:
Hello! @tanguyantoine 

#### What's this PR do?

1) It cleans up code and set correct artwork when ```updatePlayback```/```setPlayback``` called
2) Remove warning for RN 0.50
3) Fix set listener of "RNMusicControlEvent". We should set it only if it not existed. Not on each call of "on"

PR #115, opened 2017-12-06
Add stopControl method on iOS

Ayc0:
#### What's this PR do?
Proposal for issue #114

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ android/build
 *.iml
 .gradle/
 .idea/
+.vscode/
 .DS_Store
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -131,10 +131,13 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         init = true;
     }
 
-    synchronized public void destroy() {
-        if(!init) return;
+    @ReactMethod
+    public void stopControl() {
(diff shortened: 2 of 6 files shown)


tanguyantoine:
LGTM, thank you for this stuff. @Guichaguri  do you have something similar for Android ?

Ayc0:
@tanguyantoine I can do something similar for Android

tanguyantoine:
Awesome

Ayc0:
@tanguyantoine why is the code of index.android.js and index.ios.js nearly similar except for the definition of the consts at the top ?
You should have a file in which you define the const for android and ios and import those in a `index.js` file

the only real difference is `DeviceEventEmitter.addListener` for android and `new NativeEventEmitter(NativeMusicControl)` for iOS but you can detect the platform here with ReactNative.Platform === ios or === android

Ayc0:
@tanguyantoine this MR should be done

tanguyantoine:
LGTM

As discussed in #106 we would rather have a dedicated PR to merge index.ios and index.android 

@kesha-antonov started this stuff. 
@Guichaguri it sounds like everyone want to merge these to files, thoughts?

tanguyantoine:
published as 0.5.0

Guichaguri:
@tanguyantoine I'm okay with that :)

Ayc0:
I can work on a PR for the index

Guichaguri:
There is already #79, but it's outdated, so a new PR might be a good idea :)

PR #117, opened 2017-12-09
Pause on audio session interruption events on iOS

hudashot:
#### What's this PR do?

Pauses playback on iOS when an [audio interruption](https://developer.apple.com/library/content/documentation/Audio/Conceptual/AudioSessionProgrammingGuide/HandlingAudioInterruptions/HandlingAudioInterruptions.html) event is received.

None.

N/A

#### How to test:

1. Start playback.
1. Initiate a phone call to the device.
1. Observe the `pause` event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,8 @@ componentDidMount() {
       this.props.dispatch(playRemoteControl());
     })
 
-    // on iOS this event will also be triggered by the audio router change event.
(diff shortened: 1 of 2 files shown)


DanNi0130:
@hudashot Could we please add a flag to make this functionality optional? There are other libraries that implement this functionality as well that we could clash with here.

DanNi0130:
Either that or we should make the feature more fully baked (handle interruptions ending and the "should resume" flag) before merging.

PR #118, opened 2017-12-10
add end interruption should resume functionality for ios and fix typo

DanNi0130:
This pull request adds functionality to resume playback when AVAudioSessionInterruptionOptionShouldResume is set to true. Tested on iPhone 6s.


DanNi0130:
@tanguyantoine Do you care to look this over or can I just go ahead and merge it? The patch is pretty straight forward (only 7 lines)!

hudashot:
I believe you might be trying to use a previously undeclared variable here (`interruptionType` rather than `interuptionType`).

I've just proposed #119 which fixes this as well as makes interruption handling optional and disabled by default.

DanNi0130:
Closed in favor of #119.

PR #131, opened 2018-01-11
Set listener once

kesha-antonov:
#### What's this PR do?

Set listener only once. Not on each "on".


index.js
```diff
@@ -58,30 +58,29 @@ const MusicControl = {
     }
   },
   on: function(actionName, cb){
-    if(subscription){
-      subscription.remove();
+    if(listenerOfNativeMusicControl == null){
```
tanguyantoine: What about :
```js
if (!Boolean(listenerOfNativeMusicControl)){
``` 
or
```js
if (!listenerOfNativeMusicControl) {
```

index.js
```diff
@@ -58,30 +58,29 @@ const MusicControl = {
...
-        }
-      );
     handlers[actionName] = cb
   },
   off: function(actionName, cb){
     delete(handlers[actionName])
-    if(!Object.keys(handlers).length && subscription){
-      subscription.remove()
-      subscription = null;
+    if(!Object.keys(handlers).length && listenerOfNativeMusicControl != null){
```
tanguyantoine: same

index.js
```diff
@@ -58,30 +58,29 @@ const MusicControl = {
...
+      listenerOfNativeMusicControl.remove()
+      listenerOfNativeMusicControl = null;
     }
   },
   stopControl: function() {
-    if (subscription) {
-      subscription.remove();
+    if (listenerOfNativeMusicControl != null) {
```
tanguyantoine: same

index.js
Guichaguri: The second one is better, it also handles `undefined`

tanguyantoine: the first as well ;-)

kesha-antonov:
I think it's not sufficient since ```!= null``` check for null and undefined. But anyway. Changed.

tanguyantoine:
Thank you

tanguyantoine:
published as 0.5.4

kesha-antonov:
Thanks!

Issue #116, opened 2017-12-07
support lib v7 update


Nikhil-Curefit:
# Description
After updating android support lib to 27.0.2, NotificationCompat becomes deprecated and moved to v4. In v4 the issue is that NotificationCompat.MediaStyle() no longer exists.
The correct fix is to use v4 notification compat and user media style from
import android.support.v4.media.app.NotificationCompat.MediaStyle;

1. Sample code (provide repo url or sample code)
```js
//old code
nb.setStyle(new NotificationCompat.MediaStyle().setMediaSession(session.getSessionToken()));
//new code
nb.setStyle(new MediaStyle().setMediaSession(session.getSessionToken()));
```
2. Platform ?
   - [ ] iOS
   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [x] Real device

deafkittten:
I've been trying to fix this locally but I have to say that it isn't that easy. MediaStyle is not in the v4 support library. It throws all kinds of errors.

Nikhil-Curefit:
Works fine for me

deafkittten:
Do I need to add any specific support library to the dependencies?

Nekith:
Hello,

I have the same problem.

I tried adding the import, changing the setStyle call, adding "com.android.support:support-v4" dependency in react-native-music-control/android/build.gradle, and I got these errors while running gradle (via "react-native run-android"):

```
:react-native-music-control:processReleaseResources
ERROR: In <declare-styleable> FontFamilyFont, unable to find attribute android:font
ERROR: In <declare-styleable> FontFamilyFont, unable to find attribute android:fontStyle
ERROR: In <declare-styleable> FontFamilyFont, unable to find attribute android:fontWeight
```

I hope it helps somehow.
Thank you.

Nikhil-Curefit:
@deafkittten Just the support lib v4 27.0.2

Nikhil-Curefit:
@Nekith not sure why I didn't face this error. I wrote a simple script if it helps you upgrade. Just run the commands in your project folder

sed -i '' 's/android.support.v7.app.NotificationCompat;/android.support.v4.app.NotificationCompat;import android.support.v4.media.app.NotificationCompat.MediaStyle;/g' node_modules/react-native-music-control/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
[...]

PR #133, opened 2018-01-13
Upd build tools 27

kesha-antonov:
#### What's this PR do?

Update build tools to 27

Fixes https://github.com/tanguyantoine/react-native-music-control/issues/116


Guichaguri:
I don't want to update the build tools version just yet. It can cause undesirable incompatibilities with other modules.

I'll leave this PR on hold, hopefully we'll be able to update it when React Native updates its API level. https://github.com/facebook/react-native/issues/17287

kesha-antonov:
Which incompatibilities?

Wait for RN means wait forever.

Guichaguri:
A lot of other modules might use build tools dependent libraries, such as other support lib packages or Google Play Services. We can not assume that they're going to be work fine. It's better to stick with what is stable than what isn't, mainly because this project is mostly fueled by PRs and not actual collaborators.

I understand your worries that React Native will take a long time to finally update it, but they will have to do it one way or another before august, at least that's guaranteed.

kesha-antonov:
Ok! Got it.
I've updated many of modules and didn't see any problems. Although I didn't measure performance impact.

tanguyantoine:
Since Google will require minimum version to be 26 it make sense. @Guichaguri what do you think ?

jyotigupta007:
I have also faced the same issue as minimum version to be 26. So it make sense to merge this pull request and make it workable for other persons who are going to use this module in there projects. @Guichaguri  you should rethink on this.

Guichaguri:
Yes, I think we're in a point where it's safe to update it to the API level 26+, React Native is in its way to updating to the newer API levels too. Can someone please test this PR?

parveen-lukha:
Hi
As mentioned in my previous comment. "When I changed the files as Keha has done. It works fine for me."

jyotigupta007:
@Guichaguri I have tested this pull request it is working fine for me. You can merge the same.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -107,8 +108,8 @@ public void init() {
         md = new MediaMetadataCompat.Builder();
         pb = new PlaybackStateCompat.Builder();
         pb.setActions(controls);
-        nb = new NotificationCompat.Builder(context);
-        nb.setStyle(new NotificationCompat.MediaStyle().setMediaSession(session.getSessionToken()));
+        nb = new android.support.v4.app.NotificationCompat.Builder(context);
```
Guichaguri: Can you use imports instead of declaring the package name directly?

tanguyantoine:
@kesha-antonov  does that sounds doable ?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
kesha-antonov: There already imports

kesha-antonov:
@tanguyantoine I can change code but don't have time to test it.
Only few weeks later

tanguyantoine:
#157 does the trick

PR #151, opened 2018-05-05
Ios local fix

### LINKED ISSUE #146: Local Artwork Broken on iOS
# Description

Here I am again, with an identical issue that I had a while ago: Local artwork is broken on iOS using RN 0.52.

I've tracked the issue down to `MusicControlManager.m` at line 303, where it checks `BOOL fileExists = [[NSFileManager defaultManager] fileExistsAtPath:localArtworkUrl];`.  This always returns FALSE for local images, because the `uri` that RN creates is in the form `file:///var/contents...`, while NSFileManager can only handle paths in the form `/var/contents...`.  By removing the `file://` at the beginning of the `artworkUrl`, I was able to fix.

Here is code for lines `303` to `307` that works:
```
NSString *localArtworkUrl = [artworkUrl stringByReplacingOccurrencesOfString:@"file://" withString:@""];
BOOL fileExists = [[NSFileManager defaultManager] fileExistsAtPath:localArtworkUrl];
if (fileExists) {
    image = [UIImage imageNamed:localArtworkUrl];
}
```

Let me know if I should create a PR.  This is fairly simple, so might be something you can fix and test real fast @tanguyantoine without having to PR.

2. Platform ?
   - [X] iOS
   - [ ] Android
   
   
3. Device
  - [ ] Simulator
  - [X] Real device

SteffeyDev:
This fixes #146.

You can find the full details there, but in short, some change in RN broke local iOS artwork by using a new scheme for files, and this change accommodates for that and restores iOS local artwork functionality.

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -300,10 +300,10 @@ RCT_EXPORT_METHOD(observeAudioInterruptions:(BOOL) observe){
                     NSData *imageData = [NSData dataWithContentsOfURL:imageURL];
                     image = [UIImage imageWithData:imageData];
                 } else {
-                    // artwork is local. so create it from a UIImage
-                    BOOL fileExists = [[NSFileManager defaultManager] fileExistsAtPath:artworkUrl];
+                    NSString *localArtworkUrl = [artworkUrl stringByReplacingOccurrencesOfString:@"file://" withString:@""];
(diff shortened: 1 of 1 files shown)


tanguyantoine:
👍

tanguyantoine:
published as v0.5.5

Issue #153, opened 2018-05-13
Build failure on Android with SDK v27


jh3141:
Building per instructions in a project with `compileSdkVersion 27` set gives a large number of unresolved class/package errors during build.  I was able to resolve some of these issues by adding the following line to my `app/build.gradle` file's dependencies section:

```
    compile "com.android.support:support-v4:27.1.1"
```

however this leaves a number of remaining errors relating to classes in the `android.support.v4.media.session` and `android.support.v7.app` packages.  While I have not yet investigated thoroughly, it appears that the notification-related classes in the latter package at least has been deprecated for some time and may have been removed in SDK 27; see https://medium.com/google-developers/migrating-mediastyle-notifications-to-support-android-o-29c7edeca9b7 

(Unfortunately, I cannot downgrade to an older SDK because I'm also using `react-native-camera`, which requires support libs in the v27.x.x range to build)

cinder92:
There are a pull request to update build tools to 26+ please check it

tanguyantoine:
@cinder92 @Guichaguri  is #133  ok ?

tanguyantoine:
solved by #157

PR #157, opened 2018-06-04
Migrate Notifications to support Android O

renatoalencar:
#### What's this PR do?
Upgrade compatibility libs to support Android O

#153 

#### How to test:
Run with `compileSdkVersion` configured to 26 on a Android O.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -64,7 +64,7 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
 
     public NotificationClose notificationClose = NotificationClose.PAUSED;
 
-    public static final String CHANNEL_ID = "12min_playback_channel";
+    public static final String CHANNEL_ID = "react-native-music-control";
 
     public MusicControlModule(ReactApplicationContext context) {
         super(context);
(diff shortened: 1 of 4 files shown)


tanguyantoine:
we have #133 which updates to v27. any reason not to upgrade to 27 ?

renatoalencar:
Since you should keep the major version of the Android support library the same as the `targetSdkVersion`, and so the project too. Also, the default `targetSdkVersion` for React Native apps is 23, so I think that it would be more cautious to slowly increase SDK version.

Actually, there are a discussion to move React Native Android SDK version to 26: https://github.com/facebook/react-native/issues/18095.

tanguyantoine:
@Guichaguri thoughts ?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -57,6 +64,8 @@
 
     public NotificationClose notificationClose = NotificationClose.PAUSED;
 
+    public static final String CHANNEL_ID = "12min_playback_channel";
```
Guichaguri: Can you change the channel ID to something like "react-native-music-control"?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -84,6 +93,22 @@ public String getName() {
         return map;
     }
 
+    @RequiresApi(Build.VERSION_CODES.O)
+    private void createChannel(ReactApplicationContext context) {
+        NotificationManager
+            mNotificationManager =
+            (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
```
Guichaguri: There's no need to break this into 3 lines

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -84,6 +93,22 @@ public String getName() {
...
+
+        CharSequence name = "Media playback";
+        String description = "Media playback controls";
+        int importance = NotificationManager.IMPORTANCE_LOW;
```
Guichaguri: There's no need to split these into variables as they're only being used once

Guichaguri:
I agree, Google is going to enforce API level 26, and I don't think React Native is going to be updated to the latest API level. If we keep it at minimum, we won't have any problems whether they update to API 26 or update all the way to API 28.

Btw, I prefer this PR because it creates a notification channel, while #133 only updates the support lib.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
renatoalencar: Off course

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
renatoalencar: Right

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
renatoalencar: Ok, right.

Guichaguri:
Thanks

tanguyantoine:
Published as 0.7.0

Issue #159, opened 2018-06-11
0.7.0  ANR Reason: Context.startForegroundService() did not then call Service.startForeground()


MoonFMBot:
# Description

1. Sample code (provide repo url or sample code)
https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground

2. Platform ?
   - [ ] iOS
   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [x] Real device

renatoalencar:
I am facing the same problem. Actually, I think that this is my fault, on the #157 I forgot to call `startForeground` after showing the notification. But this is throwing a `cannot find symbol startForeground` error.

@tanguyantoine, have you any ideas why this is happening?

bhram:
replace this 

` ContextCompat.startForegroundService(context, new Intent(context, MusicControlNotification.NotificationService.class));
`
 into file
> MusicControlModule 

by this 
` 
context.startService(new Intent(context, MusicControlNotification.NotificationService.class));
`

and fixed

MoonFMBot:
maybe keep ContextCompat.startForegroundService  will be better option.It can prevent app enter doze mode.

tanguyantoine:
#162 should have fixed this in version 0.7.3

PR #160, opened 2018-06-13
Update MusicControlModule.java

bhram:
fix this issue https://github.com/tanguyantoine/react-native-music-control/issues/159

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -147,8 +147,8 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         receiver = new MusicControlReceiver(this, context);
         context.registerReceiver(receiver, filter);
 
-        ContextCompat.startForegroundService(context, new Intent(context, MusicControlNotification.NotificationService.class));
-
+        context.startService(new Intent(context, MusicControlNotification.NotificationService.class));
+        
         context.registerComponentCallbacks(this);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -238,7 +238,7 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
         }
 
         nb.setContentTitle(title);
-        nb.setContentText(artist);
+        nb.setContentText(artist); ContextCompat.startForegroundService(context, new Intent(context, MusicControlNotification.NotificationService.class));
```
renatoalencar: I think this would lead to the same issue described before

bhram: it's work fine for me with no issues

tanguyantoine:
@Guichaguri you are the expert :-)

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
renatoalencar: For Android Oreo too?

bhram: yep

Guichaguri: Can you explain why do you need to add this here?

Guichaguri:
Superseded by #162

PR #169, opened 2018-07-03
Using checking for value equality of artworkUrl...

cferg-entale:
... to prevent continually fetching artwork images

#### What's this PR do?
Checks for value equality of `artworkUrl` to prevent. 

The same artwork is being continually fetched when the same `artwork` is being passed to `updatePlayback:`

#### How to test:
Check that `updateArtworkIfNeeded:` is not called twice for the same URL.


tanguyantoine:
Thank you

Issue #191, opened 2018-11-10
Crashing on Android 8.0+: Fatal Exception: java.lang.IllegalStateException: Not allowed to start service Intent


jeffmon:
# Description
Getting crash reports from Firebase Crashlytics:
```
Fatal Exception: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=org.truthforlife.app/com.tanguyantoine.react.MusicControlNotification$NotificationService }: app is in background uid UidRecord{ddc63fd u0a88 CAC  bg:+45m52s470ms idle procs:1 seq(0,0,0)}
       at android.app.ContextImpl.startServiceCommon(ContextImpl.java:1538)
       at android.app.ContextImpl.startService(ContextImpl.java:1484)
       at android.content.ContextWrapper.startService(ContextWrapper.java:663)
       at android.content.ContextWrapper.startService(ContextWrapper.java:663)
       at com.tanguyantoine.react.MusicControlModule.init(MusicControlModule.java:150)
       at com.tanguyantoine.react.MusicControlModule.updatePlayback(MusicControlModule.java:291)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:160)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:789)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:29)
       at android.os.Looper.loop(Looper.java:164)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$3.run(MessageQueueThreadImpl.java:192)
       at java.lang.Thread.run(Thread.java:764)
```

More specifically with `MusicControlModule.java line 150`

This might be a possible solution? https://stackoverflow.com/questions/49961273/java-lang-illegalstateexception-not-allowed-to-start-service-intent-while-try  

1. Sample code (provide repo url or sample code)
N/A

2. Platform ?
   - [ ] iOS
   - [x] Android
   
   
3. Device
  - [ ] Simulator
  - [x] Real device

marf:
Same problem here

[...]

PR #193, opened 2018-11-12
Fixed Android 8.0+ service crash in background

marf:
#### What's this PR do?
Fix Android 8.0+ crashes
#191 
#### How to test:
Now music controls in background works and do not crash

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
@@ -1,9 +1,14 @@
 package com.tanguyantoine.react;
 
+import android.app.Notification;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.Service;
+import android.content.Context;
 import android.content.Intent;
 import android.content.res.Resources;
+import android.os.Build;
 import android.os.IBinder;
 import android.support.v4.app.NotificationManagerCompat;
 import android.support.v4.media.session.PlaybackStateCompat;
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
```diff
@@ -156,11 +175,24 @@ private int toKeyCode(long action) {
...
             return null;
         }
 
+        @Override
+        public void onCreate() {
+            super.onCreate();
+
+            if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                /*Notification notification = new Notification.Builder(MusicControlModule.INSTANCE.context, MusicControlModule.INSTANCE.CHANNEL_ID)
+                        .build();*/
```
Guichaguri: Thank you! Can you cleanup this unused code?

marf:
I have cleaned up the unused line, I think you can merge.

android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
marf: I have cleaned up the unused line, I think you can merge.

marf:
I have cleaned up the unused line, I think you can merge.

tanguyantoine:
Sorry for the delay. I published the version 0.9.0 any chance you try this one?

marf:
I will try it, but I do not think that this will cause any problem. I have read that in the new version you just supported typescript so the Android changes in the pull should not affect it.

tanguyantoine:
keep me in touch 🙏

PR #210, opened 2018-12-20
update ios state define

pengcao1:


tanguyantoine:
please add some description

tanguyantoine:
published under `0.9.5`

PR #211, opened 2018-12-20
Android cache

pengcao1:


tanguyantoine:
please add some description

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -82,6 +80,21 @@ public String getName() {
         return "MusicControlManager";
     }
 
+    private LruCache<String, Bitmap> mMemoryCache = new LruCache<String, Bitmap>(
```
Guichaguri: Please, don't use LruCache, use OkHttp's cache instead, as it takes advantage of file caching and it will share the same cache with React Native images.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -438,6 +453,15 @@ synchronized public void enableControl(String control, boolean enable, ReadableM
...
 
+    private void addBitmapToMemoryCache(String key, Bitmap bitmap) {
+        if (getBitmapFromMemCache(key) == null) {
+            mMemoryCache.put(key, bitmap);
+        }
+    }
+
+    private Bitmap getBitmapFromMemCache(String key) {
+        return mMemoryCache.get(key);
+    }
```
Guichaguri: These methods are pretty useless, they only add up to Android's method count and can be easily replaced by direct calls instead

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -273,24 +286,26 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
-                    Bitmap bitmap = loadArtwork(artworkUrl, artworkLocal);
-
-                    if(md != null) {
-                        md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, bitmap);
-                        session.setMetadata(md.build());
-                    }
-                    if(nb != null) {
-                        nb.setLargeIcon(bitmap);
-                        notification.show(nb, isPlaying);
+            if (getBitmapFromMemCache(artworkUrl) == null) {//
```
Guichaguri: You check whether the artwork has been cached, but don't use the cache at all?

Issue #206, opened 2018-12-06
Could Android lockscreen support seekbar?


pengcao1:
# Description

1. Sample code (provide repo url or sample code)
like IOS style ,so we can seek the duration to any time

2. Platform ?

   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [x] Real device

illestrater:
I'd love to see this feature supported as well!

Also even with iOS, I'm unable to seek. @pengcao1, is iOS seeking working for you? The scrubbing thumb button is not appearing for me on iOS, so I can't seem to enable seeking functionality for neither iOS or Android.

tleewu:
@illestrater I think what you're looking for is:

```
    MusicControl.enableControl('changePlaybackPosition', true)
    MusicControl.on('changePlaybackPosition', () => { console.log('hi'); })
```

I was running into the same issue, and I couldn't find this in the documentation. When I dug into the source code, I found it here: https://github.com/tanguyantoine/react-native-music-control/blob/master/ios/MusicControlManager.m#L127L129

@tanguyantoine could we add this in the README? Happy to submit a PR if it helps.

tanguyantoine:
Please do :-)

pengcao1:
how about Android?

tanguyantoine:
What do you mean? It works the same way. 
When the event is triggered then you must update your player accordingly. 
This package is about displaying information not controlling the playing status

pengcao1:
well, 
I mean that: `MusicControl.enableControl('changePlaybackPosition', true)` work fine under IOS platform. and we can touch the seek bar to change the playback position.

I don't find a prop or method in Android to enable the same function .

Guichaguri:
There's no way to implement a seek bar inside a notification, Android doesn't allow that, it's pretty restricted. Probably because the official Android design guidelines say it's not suitable to add anything other than regular buttons to the notification.

And also, why do you want to replicate the native iOS behavior on Android? Android users will expect your app to behave like any other Android music app, and not like an iOS app.

PR #215, opened 2019-01-01
Update README to include changePlaybackPosition.

tleewu:
#### What's this PR does?
Update README to include `changePlaybackPosition`, an important part to this open-source package, as it allows users to change the track position through the lockscreen.

#206


Guichaguri:
Thank you

tanguyantoine:
any change you add an example?

PR #221, opened 2019-01-11
notification not dismissing because of foreground service on android O and above

anubansal92:
#### What's this PR does?
It solves two key issues:
1. On android versions >= Oreo, **foreground service is causing the notification to become non-dismissable**, this happens only for the first notification post-initialization because from the second notification, normal NotificationManagerCompat way is used. It is a very bad experience because user won't be able to dismiss the notification unless he kills your app, which most users are unaware of, and they may incidentally end up uninstalling the app.
2. Because of foreground service, **an ugly white empty notification is created even when one is not configured** which is again non-dismissable and non-actionable because there's no way I can stop the foregroundService there.

I didn't find any issues in the issue list (because I saw that it is due to a fairly recent change), but since this is very critical for our user base, instead of creating one, raising a PR.

screenshot_20190111-182959
screenshot_20190111-182942
screenshot_20190111-182948

#### How to test:
on any android device running android O or above, follow the steps:
1. Force kill the app.
2. Open the app, play any video with NotificationClose param as "always" or "paused"
3. Now these should make the notification dismissable at least in the paused state, try dismissing the notification.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -161,12 +161,7 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
 
         Intent myIntent = new Intent(context, MusicControlNotification.NotificationService.class);
 
-        if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.O){
-            context.startForegroundService(myIntent);
-
-        }
-        else
-            context.startService(myIntent);
+        context.startService(myIntent);
 
         context.registerComponentCallbacks(this);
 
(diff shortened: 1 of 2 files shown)


anubansal92:
@tanguyantoine @Guichaguri please review this PR on priority, as this is a blocker issue for us as our app is extremely content heavy. We have a release going out on coming Monday, I'll be extremely humbled if we can resolve this before that. I'm more than happy to provide any more info or any help regarding any debugging you need.

Guichaguri:
You're basically reverting #193. You're fixing a bug by unfixing another bug.

You should be able to `MusicControl.stopControl()` when the app is closing, and that should stop the foreground service.

tanguyantoine:
@Guichaguri if that looks good to you I can publish it soon

tanguyantoine:
@anubansal92 in the meantime, please update your commit message to follow https://conventionalcommits.org `fix: your title goes here`

PR #222, opened 2019-01-16
fix some OOM \NPE and use Glide to fetch image

pengcao1:
#### What's this PR does?
1 fix the OOM issue caused by the inner Thread
2 upgrade the build.gradle for avoiding Google build warning
3 change the method of the fetched image from `Thread` to `Glide`


android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -263,39 +254,59 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
...
+                            md.putBitmap(MediaMetadataCompat.METADATA_KEY_ART, theBitmap);
+                            // session.setMetadata(md.build());
+                        }
+                        if (nb != null) {
+                            nb.setLargeIcon(theBitmap);
+                            notification.show(nb, isPlaying);
+                        }
+                    };
+                }
+            }.execute();
```
Guichaguri: You're not cancelling the task when another artwork is required, if I spam `setNowPlaying` with different artworks, I might get a random one at the end. It might be a big problem for short tracks (such as ads, jingles or even sound effects)

android/build.gradle
```diff
@@ -9,13 +9,17 @@ buildscript {
 
 apply plugin: 'com.android.library'
 
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
```
Guichaguri: I don't get it why do you need it to be configurable through extensions.
The way gradle works is that you can define the "minimum" version of everything, if the root project contains a newer version, it will override the old one. That happens both to the SDK version and to dependencies

android/build.gradle
```diff
@@ -34,6 +38,17 @@ repositories {
...
+        exclude group: "com.android.support", module: "glide"
+    }
+    implementation("com.github.bumptech.glide:annotations:${safeExtGet("glideVersion", "4.7.1")}") {
+        exclude group: "com.android.support", module: "annotations"
+    }
+    annotationProcessor "com.github.bumptech.glide:compiler:${safeExtGet("glideVersion", "4.7.1")}"
+    implementation("com.github.bumptech.glide:okhttp3-integration:${safeExtGet("glideVersion", "4.7.1")}") {
+        exclude group: "com.android.support", module: "okhttp3-integration"
+        exclude group: "com.android.support", module: "glide-parent"
+    }
```
Guichaguri: Why do you need this many Glide dependencies?

pengcao1: follow glide's official integration steps

android/build.gradle
pengcao1: first: Google will remove the `compile` option at the end of 2018, However, lots of 3rd do not upgrade them to `implementation` or `provide` etc. on time.
Second: lots of library module was used, we'd better to confirm to use the same configuration, such as miniSDK, targetSdk, and supportLibrary.

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
pengcao1: right.
could you kindly please leave your suggestion here?

android/build.gradle
Guichaguri: I'm not talking about replacing `compile` to `implementation`, I agree with that.
I'm talking about the necessity of using an extension, where you can already override a version number through Gradle's hierarchy.

pengcao1: add below code snippet to your build.gradle(the top build.gradle file)
```
ext {
    compileSdkVersion = 28
    buildToolsVersion = "27.0.3"
    supportVersion = "28.0.0"
    minSdkVersion = 19
}
```

android/build.gradle
benoitdion: We may not want to force glide onto consumers of this library, especially since React Native already ships with Fresco. 

Could we use the Fresco equivalent instead?

tanguyantoine:
closing due to inactivity

PR #225, opened 2019-01-29
fix: remove extra pause event related to app suspension in IOS10+

zzorba:
#### What's this PR does?
Apparently starting in IOS10, the OS now sends a 'Suspended' notification AFTER
an app that was suspended is resumed. The IOS implementation used to deliver this notification as a 'pause' event, which does not correspond to a user behavior. By checking for the new field on the suspended notification, I am able to suppress the false pause event from being fired to JS.

See this note here for details on the Notification change in IOS10+: https://developer.apple.com/documentation/avfoundation/avaudiosession/1616596-interruptionnotification

NA
#### How to test:
1) Setup an app with lock screen controls AND the following settings on IOS10+
```
    MusicControl.enableBackgroundMode(true);
    MusicControl.handleAudioInterruptions(true);
    MusicControl.on('play', this.play.bind(this));
    MusicControl.on('pause', this.pause.bind(this));
``
2) Play a track, trigger the lock screen, and pause the app.
3) Let the lock screen sleep (which seems to reliably trigger an app suspension for me).
4) wake the phone and select Play on the lock screen.
5) Before this change, a Play will fire, followed by an unwanted Pause event. By adding this change, teh Pause event will no longer fire, satisfying the desired user behavior.

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -370,8 +370,9 @@ RCT_EXPORT_METHOD(observeAudioInterruptions:(BOOL) observe){
     }
     NSInteger interruptionType = [notification.userInfo[AVAudioSessionInterruptionTypeKey] integerValue];
     NSInteger interruptionOption = [notification.userInfo[AVAudioSessionInterruptionOptionKey] integerValue];
-
-    if (interruptionType == AVAudioSessionInterruptionTypeBegan) {
+    bool delayedSuspendedNotification = (@available(iOS 10.0, *)) && [notification.userInfo[AVAudioSessionInterruptionWasSuspendedKey] boolValue];
+    
+    if (interruptionType == AVAudioSessionInterruptionTypeBegan && !delayedSuspendedNotification) {
         // Playback interrupted by an incoming phone call.
         [self sendEvent:@"pause"];
     }

### CODE REVIEW

File: ios/MusicControlManager.m [hunk 1]
```diff
@@ -370,8 +370,9 @@ - (void)audioInterrupted:(NSNotification *)notification {
     }
     NSInteger interruptionType = [notification.userInfo[AVAudioSessionInterruptionTypeKey] integerValue];
     NSInteger interruptionOption = [notification.userInfo[AVAudioSessionInterruptionOptionKey] integerValue];
-
-    if (interruptionType == AVAudioSessionInterruptionTypeBegan) {
+    bool delayedSuspendedNotification = (@available(iOS 10.0, *)) && [notification.userInfo[AVAudioSessionInterruptionWasSuspendedKey] integerValue] == 1;
```
tanguyantoine: do you think boolValue can work here so that we don't need to check with `1`?

tanguyantoine:
can you update your commit message by following https://www.conventionalcommits.org 🙏 otherwise LGTM

ios/MusicControlManager.m
zzorba: Absolutely, I'm a bit new to Objective-C.  Switched this over to use `boolValue`

zzorba:
Thanks for the quick review, fixed the things you mentioned (and updated the commit message)

tanguyantoine:
release under `0.9.8`

Issue #192, opened 2018-11-10
Controls hide on lock screen (Android)


ithustle:
Hi 
Its not a issue but do you have an intention to update it to show music controls on lock screen if I set notification hide on screen?

I think will be a good asset on your good work. 

Cheers 

Célio Garcia

2. Platform ?
   - [ ] iOS
   - [X ] Android
   
   
3. Device
  - [ ] Simulator
  - [X ] Real device

tanguyantoine:
If you have any skills and want to contribute, this is something we definitely can merge.

tanguyantoine:
You rock 👍

PromitRai:
did anyone manage to solve this ?

tanguyantoine:
@ithustle we are waiting your PR :-P

tanguyantoine:
haha good luck ^^

ithustle:
hehehe great incentive

tanguyantoine:
sorry, I'm not an android expert ... at all

Adam3306:
pls guys help

ithustle:
@lufinkey , I think you saved my ass 😆

lufinkey:
This issue should be re-opened since that PR crashed in some instances. I think I know a better way to implement it that doesn't hardcode the actions. I'll take another crack at it sometime this week

tanguyantoine:
reproducible with `0.10.8`?

Issue #230, opened 2019-02-08
Error: setShowActionsInCompactView


vortgo:
Hello. 
I try to use this package for my project but every time I catch error: 
**`setShowActionsInCompactView: action 0 out of bounds (max -1)`**

My code is from demo app:
```
MusicControl.enableBackgroundMode(true);
MusicControl.setNowPlaying({
      title: 'Billie Jean',
      artwork: 'https://i.imgur.com/e1cpwdo.png',
      artist: 'Michael Jackson',
      album: 'Thriller',
      genre: 'Post-disco, Rhythm and Blues, Funk, Dance-pop',
      duration: this.whoosh.getDuration(),
      description: 'Billie Jean is a song by American singer Michael Jackson. It is the second single from the singer\'s sixth solo album, Thriller (1982). It was written and composed by Jackson and produced by Jackson and Quincy Jones.',
      date: '1983-01-02T00:00:00Z',
      rating: 84
    })
    MusicControl.enableControl('play', false)
    MusicControl.enableControl('pause', true)
```

I have reduced the code to simple invoke:
```
MusicControl.setNowPlaying({
      title: 'Billie Jean',
})
```
But the error is still throwing. 
Then I have tried run demo project `https://github.com/tanguyantoine/react-native-music-control-demo` as it is and again I failed with the same error

I have found a lot of projects using this package and code is familiar with mine. 
Could any give me a hint where I make mistake?

Guichaguri:
This was caused by #227. That's what happens when you merge a commit without reviewing it or at least testing it properly.

tanguyantoine:
Can you elaborate on the reasons of this issue?

Guichaguri:
Basically, you can use `setShowActionsInCompactView` to specify which actions will be visible when the notification is collapsed. The commit in #227 hardcodes the first three actions, if you have less than those three buttons enabled, it will crash.

I don't know how a hardcoded array such as the `new int[]{0,1,2}` passes a code review.

vortgo:
I just use the previous version. Thanks for the explanation.

tanguyantoine:
Maybe because I have no knowledges on Android ;-)
I’ll revert it

PR #234, opened 2019-02-16
Android lock screen controls

lufinkey:
Fixes issues #192 and #230

Sets the android lock screen controls based on the added actions

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -112,6 +112,36 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
         mNotificationManager.createNotificationChannel(mChannel);
     }
 
+    private boolean hasControl(long control) {
+        if((controls & control) == control) {
+            return true;
+        }
+        return false;
+    }
+
+    private void updateNotificationMediaStyle() {
+        if (!(Build.MANUFACTURER.toLowerCase(Locale.getDefault()).contains("huawei") && Build.VERSION.SDK_INT < Build.VERSION_CODES.M)) {
+            MediaStyle style = new MediaStyle();
+            style.setMediaSession(session.getSessionToken());
+            int controlCount = 0;
+            if(hasControl(PlaybackStateCompat.ACTION_PLAY) || hasControl(PlaybackStateCompat.ACTION_PAUSE) || hasControl(PlaybackStateCompat.ACTION_PLAY_PAUSE)) {
+                controlCount += 1;
+            }
+            if(hasControl(PlaybackStateCompat.ACTION_SKIP_TO_NEXT)) {
+                controlCount += 1;
+            }
+            if(hasControl(PlaybackStateCompat.ACTION_SKIP_TO_PREVIOUS)) {
+                controlCount += 1;
+            }
+            int[] actions = new int[controlCount];
+            for(int i=0; i<actions.length; i++) {
+                actions[i] = i;
+            }
+            style.setShowActionsInCompactView(actions);
+            nb.setStyle(style);
+        }
+    }
+
     public void init() {
         if (init) return;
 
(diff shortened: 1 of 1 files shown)


tanguyantoine:
@Guichaguri does it look good?

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -112,6 +112,36 @@ private void createChannel(ReactApplicationContext context) {
...
+            if(hasControl(PlaybackStateCompat.ACTION_PLAY) || hasControl(PlaybackStateCompat.ACTION_PAUSE) || hasControl(PlaybackStateCompat.ACTION_PLAY_PAUSE)) {
+                controlCount += 1;
+            }
+            if(hasControl(PlaybackStateCompat.ACTION_SKIP_TO_NEXT)) {
+                controlCount += 1;
+            }
+            if(hasControl(PlaybackStateCompat.ACTION_SKIP_TO_PREVIOUS)) {
+                controlCount += 1;
+            }
+            int[] actions = new int[controlCount];
```
Guichaguri: Why not use an list instead and convert it to an array after?

lufinkey:
Because why bother? It's an unnecessary extra step?

lufinkey:
Also, I already tested this with multiple different action configurations so I know it works. I think that's all @tanguyantoine wanted you to check

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
lufinkey: Because why bother...? It's an unnecessary extra step, and I think they were just asking you to check if the commit worked.

lufinkey: Also they're still indices so you'd still need to increment controlCount, so you really wouldn't be trimming down that much code, if any.

Guichaguri: The code should be moved to `prepareNotification` from `MusicControlNotification`.
After adding the actions you can reduce the code like this:
```java
int[] actions = new int[builder.mActions.size()];
for(int i = 0; i < actions.length; i++) {
    actions[i] = i;
}
```

lufinkey: okay.
well make sure it maxes out at 3, because the lock screen notification gets really weird when there's more

lufinkey: I'm confused. Are you asking me to change this or are you changing this yourself?

lufinkey:
@Guichaguri Just an update on the changes you asked for.

Moving this code to that class requires access to the `MediaSessionCompat session` variable which is in `MusicControlModule`, because I'm required to move this code
```java
...
if (!(Build.MANUFACTURER.toLowerCase(Locale.getDefault()).contains("huawei") && Build.VERSION.SDK_INT < Build.VERSION_CODES.M)) {
    nb.setStyle(new MediaStyle().setMediaSession(session.getSessionToken()));
}
...
```
to `MusicControlNotification`, and I'm not sure if that would break anything. I have to put the action setting code with that code because it's where the MediaStyle is being set. If you know a better way to structure this, please make the PR yourself because I don't know enough about this project to start moving your code around. But in the meantime, this code works and is a feature that is very necessary for what most people are trying to do with this module.

tanguyantoine:
@Guichaguri can you do a PR on top of @lufinkey so that we know what you have in mind? :-)

13ArturBruno:
this PR resolve bug on see button in lockscreen ??

lufinkey:
Yes

13ArturBruno:
> Yes

Perfect :)

lufinkey:
Can we please merge this in?

lufinkey:
Or get it modified by @Guichaguri ?

yasahmed:
ping .... we need this changes to be merged please :'(

PR #237, opened 2019-02-24
Modernise Gradle build file

christianbach:
#### What's this PR does?
Allows gradle values to be overwritten by root project, remove obsolete compile for implements

None.
#### How to test:


Guichaguri:
Thanks

PR #238, opened 2019-02-25
Modernize Android part two

christianbach:
#### What's this PR does?
Bump gradle tools and buildToolsVersion and removes obsolete ndk options.

None.

None.

#### How to test:
Open Android studio and press Make project

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -35,6 +35,6 @@ repositories {
 }
 
 dependencies {
-    compileOnly "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
+    compileOnly 'com.facebook.react:react-native:+'
     implementation "com.android.support:support-media-compat:${safeExtGet('supportLibVersion', '26.+')}"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,9 +5,11 @@ def safeExtGet(prop, fallback) {
 buildscript {
     repositories {
         jcenter()
+        google()
     }
+
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.3.1'
+        classpath 'com.android.tools.build:gradle:3.1.4'
     }
 }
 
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -5,39 +5,36 @@ def safeExtGet(prop, fallback) {
...
-        // As RN is not updated in jcenter anymore, we'll have to grab it from npm
-        url "../node_modules/react-native/android"
-    }
     mavenCentral()
-    jcenter()
 }
 
 dependencies {
-    compileOnly 'com.facebook.react:react-native:+'
+    compileOnly "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
```
Guichaguri: Is there any specific reason for this? RN is not retrieved by a maven repository anymore, it has been moved to directly retrieve it from the `node_modules` directory, that's why it's always a plus sign.

android/src/main/java/com/tanguyantoine/react/MusicControl.java
```diff
@@ -19,6 +19,10 @@
         return modules;
     }
 
+    public List<Class<? extends JavaScriptModule>> createJSModules() {
+        return Collections.emptyList();
+    }
+
```
Guichaguri: This has been removed from RN ages ago. As you're now retrieving it from the jcenter repository instead of npm, you're getting an older version.

android/build.gradle
christianbach: Yeah, I agree. its an old pattern, I will change it right away.

android/src/main/java/com/tanguyantoine/react/MusicControl.java
christianbach: Oh, sorry! Will revert!

android/build.gradle
```diff
@@ -5,36 +5,33 @@ def safeExtGet(prop, fallback) {
...
+    lintOptions {
+        abortOnError false
     }
 }
 
 repositories {
-    maven {
-        // As RN is not updated in jcenter anymore, we'll have to grab it from npm
-        url "../node_modules/react-native/android"
-    }
```
Guichaguri: This is necessary since RN comes from npm

christianbach: Yeah, sorry about this. Reverted!

christianbach:
should be ready to go now.

Guichaguri:
Thank you

Issue #170, opened 2018-07-03
Audio interruption handling not implemented for Android


parveen-lukha:
# Description
Audio interruption on Android are not handled. Can anyone help with this?
1. Sample code (provide repo url or sample code)

2. Platform?
   - [ ] Android
   
   
3. Device
  - [ ] Simulator
  - [ ] Real device

Guichaguri:
Audio interruption must be implemented by the module that plays the audio on Android, there's nothing we can do

parveen-lukha:
Hi Guichaguri
Please refer to handleAudioInterruptions function of this module
https://github.com/tanguyantoine/react-native-music-control/blob/master/index.js

It binds NativeMusicControl.observeAudioInterruptions(enable); in case of IOS but in case of Android it just log a console
if (IS_ANDROID) {
      console.log("Audio interruption handling not implemented for Android");
}

Guichaguri:
Yes, that's because the audio interruption events in Android MUST be directly bound to a music player. As this module only implements media buttons, there's nothing we can do.

You should request this to the module that plays the audio, or use another alternative such as `react-native-track-player` which already implements it.

tanguyantoine:
Agree :-)

PR #240, opened 2019-02-26
Android interuption

roffelund:
#### What's this PR does?
- Adds handleAudioInterruptions for Android
- Separate volume logic to own class

#170


christianbach:
@Guichaguri Could you take a look at this. We are using it internally and think that this would be a great addition to this library. Since handleAudioInterruptions right now is only handled by iOS

Tejas-narola:
@tanguyantoine 
I have installed latest version 0.10.2 but we are still facing audio interruption issue on Samsung device(9.0.0).

Thanks

roffelund:
> @tanguyantoine
> I have installed latest version 0.10.2 but we are still facing audio interruption issue on Samsung device(9.0.0).
> 
> Thanks

If this is the case could you open an issue with this bug and specify what you tried, is it only android 9 you are having issues with? @Tejas-narola

PR #248, opened 2019-04-15
Add package-lock.json to source control

taylorkearns:
#### What this PR does
Adds the `package-lock.json` file to source control. I noticed it was coming up as an untracked file when I installed packages. Alternatively I could put up a PR adding `package-lock.json` to `.gitignore` if you intend to keep it out of source control, but I believe committing is the best practice.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6648 @@
+{
+  "name": "react-native-music-control",
+  "version": "0.9.9",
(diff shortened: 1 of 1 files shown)


tanguyantoine:
I would rather ignore it since we have the yarn.lock

taylorkearns:
Oh right, sorry oversight on my part. I'll open a new one ignoring the file.

PR #254, opened 2019-04-26
CloseForegroundService onDestroy

### LINKED ISSUE #247: Android - Fatal Exception: java.lang.RuntimeException: Unable to create service com.tanguyantoine.react.MusicControlNotification$NotificationService: java.util.ConcurrentModificationException
# Description

Getting this crash after updating to v0.99. It seems related to #221  which #220 supposedly fixed it. Here is the Stack trace:
image

Looking at actions user took before the crash, makes me think the controls were not actually removed when app last terminated(or device locked), since app crashes right away when trying to play a new audio. Also, it happens only on OS 8.1.0+ and on Samsung devices.

I'm calling `MusicControl.stopControl()` when the specific component handling it is closed, but when app is terminated `componentWillUnmount` is not called. Not sure if that's related.

2. Platform ?

   - [ ] iOS
   - [x ] Android
   
3. Device
  - [ ] Simulator
  - [x] Real device

roffelund:
#### What's this PR does?
Tries to fix #247 by terminating foregroundService onDestroy()

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlNotification.java
@@ -208,6 +208,11 @@ public class MusicControlNotification {
             stopSelf(); // Stop the service as we won't need it anymore
         }
 
+        @Override
+        public void onDestroy() {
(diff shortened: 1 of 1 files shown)


tanguyantoine:
Thank you 🙏

Issue #280, opened 2019-07-05
Add support for React Native 0.60, specifically androidx support libraries


timmywil:
# Description

Build a new project with `react-native init`, include react-native-music-control, and try to build in Android. The build fails with error `ERROR: Unable to resolve dependency for ':react-native-music-control@debug/compileClasspath': Could not resolve androidx.core:core:{strictly 1.0.0}.`

I've tried several workarounds, including using [jetifier](https://github.com/mikehardy/jetifier), but I think this fix will be a bit more involved than changing some paths to support libraries. Specifically, I run into issues migrating `MediaStyle` and `VolumeProviderCompat`. I'm still a novice when it comes to native Android so maybe someone with more Android experience can lend a hand.

1. Sample code (provide repo url or sample code)

https://github.com/timmywil/react-native-music-control-example

This project was created with `react-native init` and then react-native-music-control added with `yarn add react-native-music-control`. Note that RN 0.60 has autolinking, so `react-native link` should not be used.

2. Platform ?

   - [ ] iOS
   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [x] Real device

ouabing:
Add following line to `node_modules/react-native-music-control/android/build.gradle`

```
implementation 'androidx.core:core:1.0.2'
```

timmywil:
@ouabing set me on the right path, but everything needed to be moved over to androidx to compile.

ouabing:
@timmywil Using [jetifier](https://github.com/mikehardy/jetifier) and adding a postinstall script to your `package.json` will allow you to transfer non-AndroidX projects to AndroidX compatible temporally.

timmywil:
@ouabing That did not work for this project.

timmywil:
I noticed in https://github.com/react-native-community/react-native-slider they now publish a `next` tag on npm for RN 0.60. That might be a good solution here as well.

schumannd:
Does this package support RN +0.60, autolinking, androidx etc.?

tanguyantoine:
No work has been down on this side. Do you have sone link we can refer to so that we can do the necessary to make it work? Can you help in that sense?

schumannd:
@tanguyantoine 

Unfortunately I have no such experience either. But with a quick search I found [this readme](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) which seems detailed enough.

timmywil:
[...]

PR #281, opened 2019-07-08
fix(android): migrate support libraries to androidx

timmywil:
#### What's this PR does?

This fixes compilation with RN 0.60.

It's probably worth adding a note to the installation guide about not mixing the old Android support library and Androidx in the same project. Either everything needs to be upgraded to Androidx or everything needs to be downgraded to the support library. More and more people will become familiar with this process as they upgrade react native to 0.60.

I *think* this means that the release of react-native-music-control that includes this PR will require RN >= 0.60. However, it's possible that one could use the reverse [jetifier](https://github.com/mikehardy/jetifier) to convert this library back to the old support library, but I haven't tried that. Also, the jetifier alone was not enough to upgrade this library, so it may not be enough to downgrade it. I'm not sure.

Fixes gh-280

#### How to test:

Clone and checkout the `fix` branch of [this example repo](https://github.com/timmywil/react-native-music-control-example/tree/fix).

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,5 +40,6 @@ repositories {
 
 dependencies {
     compileOnly 'com.facebook.react:react-native:+'
-    implementation "com.android.support:support-media-compat:${safeExtGet('supportLibVersion', '26.+')}"
+    implementation 'androidx.core:core:1.0.2'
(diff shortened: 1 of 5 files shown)


tanguyantoine:
`0.10.5`

PR #282, opened 2019-07-09
fix(android): error can't find symbol Notification

Justkant:
#### What's this PR does?
Fix an android build error

None because the last npm release failed and I found the issue when I tried to use master directly


timmywil:
I also have this in https://github.com/tanguyantoine/react-native-music-control/pull/281

Justkant:
@timmywil I saw that but I'm not yet on 0.60 to use androidx

timmywil:
Makes sense

tanguyantoine:
thanks

tanguyantoine:
`0.10.5`

Issue #289, opened 2019-08-19
[iOS 13][Bug] Xcode 11 support


kesha-antonov:
Hey!

On iOS 13 app crashes when built with xcode 11 when calling Music Control

1. Sample code (provide repo url or sample code)

2. Platform ?

   - [x] iOS
   - [ ] Android

3. Device
  - [x] Simulator
  - [x] Real device

tanguyantoine:
Feel free to provide some logs

kesha-antonov:
```
*** Assertion failure in -[MPRemoteCommand addTarget:action:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/MediaPlayer_Sim/MobileMusicPlayer-4017.100.166/SDK/MPRemoteCommand.m:134
2019-08-19 23:02:49.616 [fatal][tid:main] Exception 'Unsupported action method signature. Must return MPRemoteCommandHandlerStatus or take a completion handler as the second argument.' was thrown while invoking enableControl on target MusicControlManager with params (
    pause,
    1,
        {
    }
)
callstack: (
	0   CoreFoundation                      0x00007fff23b60e3e __exceptionPreprocess + 350
	1   libobjc.A.dylib                     0x00007fff502f3b20 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff23b60bb8 +[NSException raise:format:arguments:] + 88
	3   Foundation                          0x00007fff2560fbb1 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 191
	4   MediaPlayer                         0x00007fff278fb210 -[MPRemoteCommand addTarget:action:] + 1154
	5   AppName                             0x000000010f905571 -[MusicControlManager toggleHandler:withSelector:enabled:] + 161
	6   AppName                             0x000000010f9036f0 -[MusicControlManager enableControl:enabled:options:] + 256
	7   CoreFoundation                      0x00007fff23b67bbc __invoking___ + 140
	8   CoreFoundation                      0x00007fff23b64fef -[NSInvocation invoke] + 287
	9   CoreFoundation                      0x00007fff23b65454 -[NSInvocation invokeWithTarget:] + 68
	10  AppName                             0x000000010f64cc32 -[RCTModuleMethod invokeWithBridge:module:arguments:] + 2658
[...]

PR #290, opened 2019-08-20
fix(ios): addTarget:action: should return MPRemoteCommandHandlerStatus

johankasperi:
#### What's this PR does?
Apple docs declares that the action given in `MPRemoteCommand` `addTarget:action` should return `MPRemoteCommandHandlerStatus` (https://developer.apple.com/documentation/mediaplayer/mpremotecommand/1622895-addtarget?language=objc)

The current implementation of the actions returns `void` and this causes crashes on iOS 13.

This PR returns `MPRemoteCommandHandlerStatusSuccess` in all actions. 

https://github.com/tanguyantoine/react-native-music-control/issues/289

#### How to test:
1. Call `enableControl` on an iOS 13 device will result in a crash
2. Apply this PR
3. Call `enableControl` on an iOS 13 device will result in no crash

### DIFF
--- a/ios/MusicControlManager.m
+++ b/ios/MusicControlManager.m
@@ -264,19 +264,57 @@ RCT_EXPORT_METHOD(observeAudioInterruptions:(BOOL) observe){
     [[NSNotificationCenter defaultCenter] removeObserver:self name:AVAudioSessionRouteChangeNotification object:nil];
 }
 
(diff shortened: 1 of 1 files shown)


kesha-antonov:
Nice!

tanguyantoine:
Thank you

PR #311, opened 2019-12-09
Fix the app crash when run on android api level <= 23

danielle-Cman:
Adapt run all android api level

#### What's this PR does?
Fix the app crash when run on android api level <= 23
Adapt run all android api level.

related in function
```
private void updateNotificationMediaStyle()
```
The current code:
```
 if (!(Build.MANUFACTURER.toLowerCase(Locale.getDefault()).contains("huawei") && Build.VERSION.SDK_INT < Build.VERSION_CODES.M)) {
```
We should compare android device have no `huawei` and `Build.VERSION.SDK_INT > Build.VERSION_CODES.M`
```
 if (!Build.MANUFACTURER.toLowerCase(Locale.getDefault()).contains("huawei") && Build.VERSION.SDK_INT > Build.VERSION_CODES.M) {
```
Because i test on emulator and the app crash so i cant take a screenshots.
#### How to test:
We could run on emluator with android api level > 23 and <= 23.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -118,7 +118,7 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     }
 
     private void updateNotificationMediaStyle() {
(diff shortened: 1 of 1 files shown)


tanguyantoine:
thank you

tanguyantoine:
published under `0.10.8`

danielle-Cman:
Thank you!

ouabing:
This change just disabled notification media style for all the Huawei devices and also all the Android 6.0 devices. Shall we revert it?

Issue #244, opened 2019-03-18
(Android 9 / Samsung S9) Unable to create service com.tanguyantoine.react.MusicControlNotification$NotificationService:


jamsch:
# Description

App crashes with the following devices in a production app:

- SM-G960U
- SM-G960F
- SM-G960W

All devices listed above are running Android 9

2. Platform ?
    - [ ] iOS
    - [x] Android
  
3. Device
    - [ ] Simulator
    - [x] Real device

## Stack trace

![Selection_029](https://user-images.githubusercontent.com/12927717/54565340-d9898780-4a32-11e9-900d-6a3f0ca410cb.png)

```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.app.Notification com.tanguyantoine.react.MusicControlNotification.a(android.support.v4.app.y$c, boolean)' on a null object reference
    at com.tanguyantoine.react.MusicControlNotification$NotificationService.onCreate
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:3728)
    at android.app.ActivityThread.access$1400(ActivityThread.java:235)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1784)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:214)
    at android.app.ActivityThread.main(ActivityThread.java:6981)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1445)
java.lang.RuntimeException: Unable to create service com.tanguyantoine.react.MusicControlNotification$NotificationService: java.lang.NullPointerException: Attempt to invoke virtual method 'android.app.Notification com.tanguyantoine.react.MusicControlNotification.a(android.support.v4.app.y$c, boolean)' on a null object reference
    at android.app.ActivityThread.handleCreateService(ActivityThread.java:3740)
    at android.app.ActivityThread.access$1400(ActivityThread.java:235)
    at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1784)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:214)
    at android.app.ActivityThread.main(ActivityThread.java:6981)
    at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #316, opened 2020-01-29
Stop Foreground Service directly instead of starting and then stoping in onStartCommand.

kalpeshp0310:
Stop Foreground Service directly instead of starting the service and then stopping in onStartCommand.

#### What's this PR does?
Tries to fix part of crash mentioned in #244 

As per crash log App crashes in `onCreate` method of `NotificationService`.

And as per logs from our App, Service was started right after callback on `onAudioFocusChange` with `AudioManager.AUDIOFOCUS_LOSS`.

This means that, App is crashing only when we try to stop foreground service.

So, instead of starting service and then stopping in onStartCommand. We can stop service by directly calling stop service. Same as how we stop service in `MusicControlNotification::hide` method.

NA
#### How to test:
Unfortunately I do not reproduction steps.

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlEventEmitter.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlEventEmitter.java
@@ -2,12 +2,15 @@ package com.tanguyantoine.react;
 
 import android.content.Intent;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


tanguyantoine:
`0.11.0`

PR #336, opened 2020-06-22
Android Fixes

SBShane:
This PR fixes 3 Android issues I came across.

1. The skipBackward and skipForward intervals only worked the first time a notification was created. If controls were stopped with stopControl() and then re-created, the intervals didn't take effect and would only show the 10 second icon. They have to be re-applied during the init method

2. If a color value was supplied, then there was a "removeFade" logic that reset the media style. Unfortunately, this prevented actions from appearing in compact view and also prevented the seek bar from appearing (Android 10+ only). A better workaround seems to be to instead use NotificationCompat.Builder's setColorized method to prevent Android from using colors it extracts from the artwork. For optimal flexibility, I added a "colorized" property which allows a few different scenarios:
  - Setting a color value, but not isColorized will result in the color showing for all versions of Android
  - Setting a color value and isColorized=false will also result in the color showing for all versions
  - Setting a color value and isColorized=true will result in the color showing on Android < 8.0, while still allowing Android 8.0+ to extract the color from the artwork
  - If neither the color nor isColorized values are provided, then Android < 8.0 will default to the grey color, while Android 8.0+ will extract the color from the artwork

3. The seek bar is supported in Android 10+, but didn't auto-increment during playback if a speed value wasn't explicitly supplied. In that case the speed would default to 0, which is invalid. The default speed should instead be 1, indicating normal playback

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ MusicControl.setNowPlaying({
   genre: 'Post-disco, Rhythm and Blues, Funk, Dance-pop',
   duration: 294, // (Seconds)
   description: '', // Android Only
-  color: 0xFFFFFF, // Notification Color - Android Only
+  color: 0xFFFFFF, // Android Only - Notification Color
+  colorized: true, // Android 8+ Only - Notification Color extracted from the artwork. Set to false to use the color property instead
   date: '1983-01-02T00:00:00Z', // Release Date (RFC 3339) - Android Only
   rating: 84, // Android Only (Boolean or Number depending on the type)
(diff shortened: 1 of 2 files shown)


bradleyflood:
Thanks @SBShane this looks good. Let me test and review then I will get back to you.

SBShane:
Latest commit is simply in the interest of being complete, but does not change the functionality of the player in any way. I mentioned that a speed value of 0 is invalid which is only partially true. The speed can be 0 when the player is paused, but should never be 0 while the player is playing.

PR #337, opened 2020-06-23
[Android] Add setNotificationIds function to MusicControl

hellostu:
#### What's this PR does?
Add a new Android-only function to MusicControl: `setNotificationIds(notificationId: number, channelName: String)`. Before these were statically set in the source-code, however it would be useful to set these from within the js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,12 @@ Resets, hides the music controls and disables everything.
 MusicControl.stopControl()
 ```
 
+### Set notification id and channel id (Android Only). 
+```javascript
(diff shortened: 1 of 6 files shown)


hellostu:
@bradleyflood did a new PR as the last PR wasn't working properly. This adds the functionality I need, would be good to get it into the lib if you think it's beneficial. Cheers!

bradleyflood:
Hi @hellostu sorry I've taken so long to get to this - I've been really under the pump with my own app.

I'll get to this soon - it's on my todo list :) Thanks mate.

PR #354, opened 2020-10-14
fixed hide method to work with older SDKs (Android)

frknnay:
#### What's this PR does?
It fixes the problem that the 'hide' method doesn't work with SDKs older than 26 (Android).

It is not related to any issues.

#### How to test:
With an Android SDK older than 26, try to enable the music controller then stop it. You'll see that controls didn't hide.


bradleyflood (approved):
LGTM

Issue #358, opened 2020-11-17
TypeScript Error: MusicControl.on expecting enum


redreceipt:
# Description

1. Sample code (provide repo url or sample code)
```
    MusicControl.on('play', () => {
      setIsPlaying(true);
    });
```

produces this error...

<img width="1522" alt="Screen Shot 2020-11-17 at 9 47 38 AM" src="https://user-images.githubusercontent.com/2659478/99406195-9e042800-28bb-11eb-9342-a4a1ec019a89.png">

The `Command` enum doesn't seem to be exported so there's no way to use the expected parameter. Strings seem to still work fine.

2. Platform ?

   - [x] iOS
   - [x] Android
   
   
3. Device
  - [x] Simulator
  - [x] Real device

PR #359, opened 2020-11-17
Makes Command ENUM Optional

redreceipt:
This will allow strings to still work on the `MusicControl.on` function.

#358 

Screen Shot 2020-11-17 at 9 47 38 AM

#### How to test:

Pass a string and watch the ts error go away

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -53,7 +53,7 @@ const MusicControl = {
   enableControl: function(controlName: string, enable: boolean, options = {}) {
     NativeMusicControl.enableControl(controlName, enable, options || {})
   },
-  handleCommand: function(commandName: Command, value: any) {
+  handleCommand: function(commandName: Command | string, value: any) {
     if (handlers[commandName]) {
       //@ts-ignore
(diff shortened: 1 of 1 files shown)


bradleyflood (approved):

bradleyflood:
Thanks @redreceipt

tanguyantoine:
Shouldn’t  we export the enum? That way no more magic string. Wdyt?

bradleyflood:
Yes you are right @tanguyantoine . I was just being lazy (time poor), but we should do it the right way.

So I'm not familiar with Typescript (Flow user), but I can see that `Command` is exported from `src/types.ts` as an Object, so that does object need to be imported and referenced for all `.on()` arguments? If so shouldn't it be as `.js`?

redreceipt:
@tanguyantoine @bradleyflood should be as simple as exporting `Command` from `index.ts` right? So we can import and use the function like this `MusicControl.on(Command.play, () => null)`

tanguyantoine:
yup

PR #363, opened 2020-11-23
Fix Xcode 12 build errors

dylancom:
This affects projects that make use of use_frameworks! in their podspec.
It seems that many projects we're using the wrong dependency.


tanguyantoine (approved):
LGTM @bradleyflood  any thoughts?

dylancom:
@tanguyantoine for more context, checkout: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

PR #381, opened 2021-03-29
exception handling added and crash fixed

krunalbad:
**Update msg for** 

Commit 1: exception handling added for the regular crashes on these lines when app is background for longer period or if get conflicted with other channels/pkgs

Commit 2: reduces crashes like ActivityThread.java and also fixes the crash when killing the service or app

### DIFF
--- a/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
+++ b/android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
@@ -322,6 +322,7 @@ public class MusicControlModule extends ReactContextBaseJavaModule implements Co
     @ReactMethod
     synchronized public void setNowPlaying(ReadableMap metadata) {
         init();
+        if (notification == null) return;
         if(artworkThread != null && artworkThread.isAlive()) artworkThread.interrupt();
         artworkThread = null;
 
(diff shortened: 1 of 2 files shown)


bradleyflood (approved):

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
```diff
@@ -369,8 +370,10 @@ synchronized public void setNowPlaying(ReadableMap metadata) {
         nb.setContentInfo(album);
         nb.setColor(notificationColor);
         nb.setColorized(false);
-
-        notification.setCustomNotificationIcon(notificationIcon);
+ 
+        if(notificationIcon != null){
```
bradleyflood, line 374: You've already added `if (notification == null) return;` at the start of this function (line 325), so this check here is not unnecessary, right?

bradleyflood (changes requested):

android/src/main/java/com/tanguyantoine/react/MusicControlModule.java
krunalbad: this is for `notificationIcon`. so if it is null then we should not even let it execute unless it is a URL found from setnowplaying. so I think this should be good.

bradleyflood (approved):

PR #382, opened 2021-04-19
Adding mavenCentral() as jcenter() is shutting

maheshwarimrinal:
Adding mavenCentral() as jcenter() is shutting down

#### What's this PR does?

Adding mavenCentral() and removing jcenter() from build.gradle


bradleyflood (approved):

bradleyflood:
Thanks @maheshwarimrinal :)

PR #385, opened 2021-05-11
Removed extra parenthesis from README example

srobbin:
#### What's this PR does?

It removes an extra set of parenthesis from the skipBackward and skipForward examples in the README file, and provides object formatting to match the other examples.


bradleyflood (approved):
Thanks!

PR #387, opened 2021-05-28
Just a simple change in readme

krunalbad:


bradleyflood:
Hi @krunalbad thanks for the PR. I don't think its something we need right now, but appreciate the offer. 
Thank you.

PR #389, opened 2021-06-03
Foreground crash fix in android

krunalbad:
```
Context.startForegroundService() did not then call Service.startForeground(): 
ServiceRecord{360f674 u0 com.project.bundleid/com.tanguyantoine.react.MusicControlNotification$NotificationService}
```

@bradleyflood Please review this in your availability and if it is proper then merge otherwise close it...I am not sure if this might fix the crash...It will still take a couple of days for me to test it thoroughly


krunalbad:
@bradleyflood I am going to close this. since this doesn't resolve my issue

nahuelb:
@krunalbad were you able to fix that error any other way?

krunalbad:
no. due to some other priority tasks. I am still letting this issue bother me. still haven't found the solution yet. in fact, the exception handling doesn't do the work. it's something related to inside android that is having conflicts with the notification channel

